PR #4, opened 2026-02-03
Add CMakeLists.txt for tests

mrtcnk:
Adds a missing CMakeLists.txt under tests/ to allow building the test targets with CMake.


tests/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+add_executable(test_equality_proof test_equality_proof.c)
+target_link_libraries(test_equality_proof PRIVATE mpt-crypto)
+add_test(NAME EqualityProofTests COMMAND test_equality_proof)
+
+# Add the Same Plaintext Proof tests
+add_executable(test_same_plaintext test_same_plaintext.c)
+target_link_libraries(test_same_plaintext PRIVATE mpt-crypto)
+add_test(NAME SamePlaintextProofTests COMMAND test_same_plaintext)
+
+
```
mathbunnyru, line 34: Let's not have double empty lines in this file, doesn't make sense to me (please check all places in this file)

tests/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+add_executable(test_same_plaintext_multi test_same_plaintext_multi.c)
+target_link_libraries(test_same_plaintext_multi PRIVATE mpt-crypto)
+add_test(NAME SamePlaintextMultiTests COMMAND test_same_plaintext_multi)
+
+# Add the test executable
+add_executable(test_link_proof tests test_link_proof.c)
+target_link_libraries(test_link_proof mpt-crypto secp256k1 OpenSSL::Crypto)
+add_test(NAME LinkProofTests COMMAND test_link_proof)
+
+#ElGamal Reveal Verification (The one we just wrote)
```
mathbunnyru: Let's have a space between hash and comment (also check all places in this file)

mathbunnyru (changes requested):

tests/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
+add_executable(test_elgamal test_elgamal.c)
+
+# Link the test against our library and its dependencies
+target_link_libraries(test_elgamal PRIVATE mpt-crypto)
+
+# Add the test to CTest
+add_test(NAME ElGamalTests COMMAND test_elgamal)
```
mathbunnyru: Now I see its the same as next declarations, could you please create a function which does 3 things: add_executable + target_link_libraries + add_test

And this way you'll just call this function for all the tests you need

mathbunnyru (changes requested):

tests/CMakeLists.txt
mrtcnk: I have updated by including a function.

mrtcnk (commented):
The function is included.

mathbunnyru (approved):

PR #10, opened 2026-02-09
Add cmake-build-* to gitignore

mathbunnyru:


bthomee (approved):

bthomee:
@mathbunnyru it would perhaps be easier to just use the same .gitignore as the rippled repo?

mathbunnyru:
> @mathbunnyru it would perhaps be easier to just use the same .gitignore as the rippled repo?

It has some rippled-related things, so let's keep it simple here for now.

PR #11, opened 2026-02-10
Fix tests failing in release build

mrtcnk:
Tests: Replaced assert() in all test files with a custom persistent EXPECT() macro. This guarantees that cryptographic setup logic executes correctly in both Debug and Release modes.

CMake: Moved architecture detection into CMakeLists.txt using CheckTypeSize. It now automatically detects 128-bit support and applies the optimized 64-bit flags (USE_SCALAR_4X64) globally.

Cleanup: Removed manual #define blocks from src/mpt_scalar.c to rely entirely on CMake, preventing "macro redefinition" warnings.

Verification I have verified that all tests now pass successfully in both Debug and Release modes on macOS.


tests/test_bulletproof.c
```diff
@@ -1,107 +0,0 @@
-#include <stdio.h>
```
mathbunnyru, line 1: Why was this test deleted?

tests/test_bulletproof_agg.c
```diff
@@ -15,45 +15,14 @@
...
-
-extern int secp256k1_mpt_get_generator_vector(
-        const secp256k1_context* ctx,
-        secp256k1_pubkey* vec,
-        size_t n,
-        const unsigned char* label,
-        size_t label_len
-);
+/* --- Macro: Persistent Assertion --- */
+#define EXPECT(cond, msg) do { \
```
mathbunnyru: I saw this macro being used in another place, could you please not duplicate code?
You can put reused macros/functions in a header and include it

tests/test_commitments.c
```diff
@@ -1,18 +1,24 @@
 #include <stdio.h>
 #include <string.h>
-#include <assert.h>
+#include <stdlib.h> // Required for abort()
```
mathbunnyru: These types of comments are rarely needed, and they are not used uniformly, let's remove it (in all files you changed)

CMakeLists.txt
```diff
@@ -1,7 +1,20 @@
 cmake_minimum_required(VERSION 3.16)
 project(mpt-crypto C CXX)
 
-# --- Find Dependencies ---
+# --- 1. Global Architecture Detection ---
```
mathbunnyru: I would remove `1`, since there is no `2`

CMakeLists.txt
```diff
@@ -1,7 +1,20 @@
...
+include(CheckTypeSize)
+check_type_size("unsigned __int128" HAVE_INT128_T)
+
+if (HAVE_INT128_T OR CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "arm64"
+    OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+    message(STATUS "Build: Detected 64-bit system. Enforcing 64-bit arithmetic globally.")
+    add_compile_definitions(USE_SCALAR_4X64 USE_FIELD_5X52 HAVE___INT128)
+else ()
+    message(STATUS "Build: Detected 32-bit system. Enforcing generic arithmetic globally.")
+    add_compile_definitions(USE_SCALAR_8X32 USE_FIELD_10X26)
```
mathbunnyru: Instead of add_compile_definitions, it's usually better to use target_compile_definitions, can we use it here?

CMakeLists.txt
```diff
@@ -1,7 +1,20 @@
...
+endif ()
+
+# --- Find Dependencies (Maintainer's Way) ---
```
mathbunnyru: It's not `Maintainer's Way`

mathbunnyru (changes requested):

src/mpt_scalar.c
```diff
@@ -36,15 +36,8 @@
...
-/* 1. Backend Configuration Definitions */
-#ifndef USE_SCALAR_8X32
-#define USE_SCALAR_8X32
-#endif
-#ifndef USE_FIELD_10X26
-#define USE_FIELD_10X26
-#endif
 
-/* 2. Include low-level utilities first.
+/* Include low-level utilities first.
```
mathbunnyru, line 40: Next block starts with `3.`, which doesn't make sense now

mathbunnyru:
Also, the PR title doesn't represent the change, since it's not Conan-related at all, but tests failing in Release builds.
I updated the title

tests/test_bulletproof.c
mrtcnk: Thank you for the review. I removed test_bulletproof.c because it wasn't referenced in tests/CMakeLists.txt, so it wasn't being compiled or run. Since test_bulletproof_agg.c covers the Bulletproof functionality (and generalizes the single-proof case), I thought it was cleaner to remove the redundant file rather than maintain it.

tests/test_bulletproof_agg.c
mrtcnk: I’ve refactored the code to use a shared tests/test_utils.h header. I moved the EXPECT macro there and updated all test files to include it.

CMakeLists.txt
mrtcnk: I’ve updated CMakeLists.txt to use target_compile_definitions with the PUBLIC keyword.

CMakeLists.txt
```diff
@@ -1,6 +1,20 @@
...
 
+# --- Global Architecture Detection ---
+include(CheckTypeSize)
+check_type_size("unsigned __int128" HAVE_INT128_T)
+
+# Store definitions in a variable
+if (HAVE_INT128_T OR CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "arm64"
+    OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+    message(STATUS "Build: Detected 64-bit system. Enforcing 64-bit arithmetic.")
+    set(MPT_ARCH_DEFS USE_SCALAR_4X64 USE_FIELD_5X52 HAVE___INT128)
```
mathbunnyru: I haven't found these defines in this library or secp256k1 library. We need to figure out how to properly set some defines, if they need to be set at all

tests/test_bulletproof.c
mathbunnyru: Got it, thanks 👍

tests/test_link_proof.c
```diff
@@ -1,11 +1,21 @@
...
 #include <string.h>
-#include <assert.h>
+#include <stdlib.h> // Required for abort()
 #include "secp256k1_mpt.h"
+#include <secp256k1.h>
 #include <openssl/rand.h>
+#include "test_utils.h"
+
+/* --- Helper: Safe Random Scalar Generation --- */
+static void random_scalar(const secp256k1_context* ctx, unsigned char* out) {
```
mathbunnyru: This is also duplicated.
Could you please move all the code which is used more than 1 time to your test_utils.h?

CMakeLists.txt
```diff
@@ -1,6 +1,19 @@
...
+# Check if the compiler supports __int128 (required for the optimized 64-bit code)
+check_type_size("__int128" HAVE___INT128)
+
+if (HAVE___INT128)
+    message(STATUS "Build: Detected 128-bit integer support. Enabling 64-bit optimizations.")
+    set(MPT_ARCH_DEFS SECP256K1_WIDEMUL_INT128 USE_FIELD_5X52 HAVE___INT128)
+else ()
+    message(STATUS "Build: No 128-bit support detected. Using generic 32-bit arithmetic.")
+    set(MPT_ARCH_DEFS USE_SCALAR_8X32 USE_FIELD_10X26)
+endif ()
```
mathbunnyru: Please, check for every macro you define, if they are actually present in this or sec256k1 project.

mrtcnk: I've updated CMakeLists.txt to use the definitions that actually exist in the source. 
64-bit/Int128: Sets SECP256K1_WIDEMUL_INT128 and USE_FIELD_5X52.
32-bit/Fallback: Sets SECP256K1_WIDEMUL_INT64 and USE_FIELD_10X26.
This now correctly triggers the optimized arithmetic paths.

tests/test_link_proof.c
mrtcnk: I have refactored the duplicated logic (specifically the RNG helpers and context setup) into test_utils.h and updated all relevant test files to use these shared functions.

CMakeLists.txt
mrtcnk: I've updated CMakeLists.txt to use the definitions that actually exist in the source.

tests/test_same_plaintext_multi_shared_r.c
```diff
@@ -1,139 +1,126 @@
...
-    /* 1. Setup Context */
-    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY | SECP256K1_CONTEXT_SIGN);
     unsigned char seed[32];
-    if (RAND_bytes(seed, 32) == 1) {
-        secp256k1_context_randomize(ctx, seed);
-    }
+    FILE *fr = fopen("/dev/urandom", "r");
+    EXPECT(fr != NULL, "Failed to open /dev/urandom");
+    EXPECT(fread(seed, 1, 32, fr) == 32, "Failed to read random seed");
+    fclose(fr);
```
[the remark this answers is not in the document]
mrtcnk: Fixed. I removed the non-portable /dev/urandom code and replaced it with the random_bytes() helper (wrapping RAND_bytes) to ensure cross-platform compatibility.

tests/test_same_plaintext_multi_shared_r.c
```diff
@@ -1,139 +1,126 @@
...
-    assert(res == 0);
-    printf("Tamper detection: OK.\n");
 
-    /* 8. Negative Test: Tamper with Transaction Context */
-    printf("Verifying proof with wrong TxID (Expecting Failure)...\n");
-    unsigned char tx_id_fake[32];
-    memcpy(tx_id_fake, tx_id, 32);
-    tx_id_fake[0] ^= 0xFF; // Flip bits
+    EXPECT(res == 1, "Proof Verification Function returned failure");
+    printf("Proof verified successfully.\n");
```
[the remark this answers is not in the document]
mrtcnk: Fixed. I have reintroduced negative test cases.

tests/test_same_plaintext_multi_shared_r.c
```diff
@@ -1,139 +1,126 @@
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
-#include <assert.h>
-#include <openssl/rand.h>
-#include <stdlib.h> // Added for malloc/free
-
+#include <time.h>
```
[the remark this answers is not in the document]
mrtcnk: Fixed.

tests/test_same_plaintext.c
```diff
@@ -1,18 +1,43 @@
...
+    } \
+} while(0)
+
+/* Helper to dump pubkey bytes for debugging */
+static void dump_pubkey_raw(const char* name, const secp256k1_pubkey* pk) {
+    const unsigned char* p = (const unsigned char*)pk;
+    printf("%s raw: ", name);
+    for (size_t i = 0; i < sizeof(*pk); i++) printf("%02x", p[i]);
+    printf("\n");
+}
```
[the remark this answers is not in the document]
mrtcnk: Fixed. Removed the unused dump_pubkey_raw function.

tests/test_same_plaintext_multi_shared_r.c
```diff
@@ -1,139 +1,126 @@
...
+    int res = secp256k1_mpt_prove_equality_shared_r(
             ctx, proof,
+            amount,          // correct arg order
+            r,
             N_RECIPIENTS,
-            &C1, C2, Pk, tx_id
+            &C1,
+            C2s,             // Pass the array directly
+            pks,             // Pass the array directly
+            tx_context
```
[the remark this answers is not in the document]
mrtcnk: Fixed. Updated proof buffers to use the _size() API helper instead of hard-coded values.

CMakeLists.txt
```diff
@@ -1,7 +1,20 @@
...
+# --- 1. Global Architecture Detection ---
+include(CheckTypeSize)
+check_type_size("unsigned __int128" HAVE_INT128_T)
+
+if (HAVE_INT128_T OR CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "arm64"
+    OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+    message(STATUS "Build: Detected 64-bit system. Enforcing 64-bit arithmetic globally.")
+    add_compile_definitions(USE_SCALAR_4X64 USE_FIELD_5X52 HAVE___INT128)
+else ()
+    message(STATUS "Build: Detected 32-bit system. Enforcing generic arithmetic globally.")
```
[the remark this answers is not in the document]
mrtcnk: I've updated CMakeLists.txt to use the definitions that actually exist in the source.

CMakeLists.txt
```diff
@@ -22,8 +35,8 @@ add_library(mpt-crypto
 target_include_directories(mpt-crypto PUBLIC include)
 
 # --- Set Compile Definitions ---
-target_compile_definitions(mpt-crypto PRIVATE USE_SCALAR_8X32 USE_FIELD_10X26 ECMULT_WINDOW_SIZE=15
-                                              ECMULT_GEN_PREC_BITS=4)
+# Pass the architecture flags (MPT_ARCH_DEFS) to the compiler
+target_compile_definitions(mpt-crypto PRIVATE ECMULT_WINDOW_SIZE=15 ECMULT_GEN_PREC_BITS=4 PUBLIC ${MPT_ARCH_DEFS})
```
mathbunnyru: `ECMULT_GEN_PREC_BITS` - also doesn't exist?

mathbunnyru: I would also remove `ECMULT_WINDOW_SIZE`, since 15 is the default: https://github.com/bitcoin-core/secp256k1/blob/master/CMakeLists.txt#L61

mrtcnk: Fixed.

mathbunnyru:
I simplified cmake further - we don't need to have a separate variable, if move the section further down
And we can also make SECP256K1_WIDEMUL_INT128/SECP256K1_WIDEMUL_INT64 private definition, since we don't expose it as part of the library

mrtcnk:
Thanks, that looks great! Moving the logic down and using PRIVATE definitions is much cleaner. I appreciate the simplification!

mathbunnyru (approved):

PR #12, opened 2026-02-17
Add SDK utility for off-chain operations

yinyiqian1:
All the off-chain operations should go through this layer.
Interfaces include:
- Key generation
- Blinding factor generation
- Encryption
- Decryption
- Context hash generation for each tx
- Proof generation for each tx
- Proof verification for each tx

The interface can be found in:
- `include/utility/mpt_utility.h`

The implementation details:
- `src/utility/mpt_utility.cpp`

For details on how to use these functions, the test cases contain the demos on how to generate proof for each transaction; it also includes how to verify the proof for each transaction:
- `tests/test_mpt_utility.cpp`

It acts as the producer of the required off-chain values when submitting requests to rippled, providing a single source of truth for all off-chain operations so everyone can consistently prepare their requests.

This can be consumed by:
- Rippled's test framework
- QA framework
- Libs
- Other users


include/utility/mpt_utility.h
```diff
@@ -0,0 +1,376 @@
...
+#include <cstddef>
+#include <secp256k1.h>
+#include <stdint.h>
+#include <vector>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// XRPL Transaction Types, the number MUST match rippled's definitions
```
mathbunnyru, line 13: Let's use the common naming style for these variables, and make them start with common prefix, for example `size_half_sha` -> `kCONFIDENTIAL_MPT_HALF_SHA_SIZE`

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,376 @@
...
+    uint8_t const pub[size_pubkey],
+    uint8_t const context_hash[size_half_sha],
+    uint64_t const amount,
+    uint8_t const encrypted_amount[size_gamal_ciphertext_total],
+    uint8_t out_proof[size_equality_proof]);
+
+#ifdef __cplusplus
+}
+#endif
+#endif
```
mathbunnyru: Please, make sure you pass pre-commit checks

tests/test_mpt_utility.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    uint8_t priv[size_privkey];
+    uint8_t pub[size_pubkey];
+    uint8_t bf[size_blinding_factor];
+    uint8_t ciphertext[size_gamal_ciphertext_total];
+    assert(mpt_generate_keypair(priv, pub) == 0);
+
+    std::vector<uint64_t> test_amounts = {
+        0, 1, 1000,
+        // 123456789,          // lib bug large numbers can not pass
```
mathbunnyru: Let's improve a comment, it's not very clear

mathbunnyru: @mrtcnk is this a problem that there is such a bug?

tests/test_mpt_utility.cpp
```diff
@@ -0,0 +1,367 @@
...
+        // 10000000000ULL
+    };
+
+    for (uint64_t original_amount : test_amounts)
+    {
+        uint64_t decrypted_amount = 0;
+
+        assert(mpt_generate_blinding_factor(bf) == 0);
```
mathbunnyru: We don't use `assert`s, and I also think this is why CI is red

.clang-format
```diff
@@ -0,0 +1,83 @@
+---
```
mathbunnyru, line 1: Please, make a separate PR adding clang-format (and a pre-commit hook)

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,641 @@
...
+
+/**
+ * Context for secp256k1 operations.\
+ * Initialized once and reused across all operations to optimize performance
+ */
+secp256k1_context*
+mpt_secp256k1_context()
+{
+    struct ContextHolder
+    {
```
mathbunnyru, line 40: I would suggest moving this struct out of this function, it will make it easier to read this function (as it will only have 2 lines)

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,641 @@
...
+#define htobe64(x) __builtin_bswap64(x)
+#define be64toh(x) __builtin_bswap64(x)
+#else
+#define htobe64(x) (x)
+#define be64toh(x) (x)
+#endif
+#endif
+
+/**
+ * Context for secp256k1 operations.\
```
mathbunnyru:
```suggestion
 * Context for secp256k1 operations.
```

mathbunnyru (changes requested):

src/utility/mpt_utility.cpp
yinyiqian1: I prefer keeping it encapsulated, because nowhere else should ever use it. It does not need to be visible to the rest of the file. No one can accidentally try to create their own ContextHolder.

mathbunnyru: Ok, let's keep it like that

include/utility/mpt_utility.h
mathbunnyru: Don't you want to do the same for `ttCONFIDENTIAL_MPT_CONVERT`-like vars?

yinyiqian1: prefix `kCONFIDENTIAL_MPT_....` seems too long if I add it to the size constants. So I use `kMPT_...`

`ttCONFIDENTIAL_MPT_CONVERT` is the same name as rippled, so I keep the `ttCONFIDENTIAL_MPT` prefix, tt indicates it is a transaction name

tests/test_mpt_utility.cpp
yinyiqian1: @mrtcnk This is the known issue, that is for decrypting large numbers. Since our for loop is limited, so they are not supported yet.

mrtcnk: Yes, this limitation is intentional for performance reasons at the moment. A more efficient approach will be introduced soon.

mrtcnk: Maybe we can improve or remove this comment to make things clearer.

include/utility/mpt_utility.h
mathbunnyru: Got it, thanks

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
+#ifndef MPT_UTILITY_H
+#define MPT_UTILITY_H
+
+#include <cstddef>
+#include <secp256k1.h>
+#include <stdint.h>
+#include <vector>
```
mathbunnyru: This header is not meant to be used in C code, right?
Then, instead of all the defines, we should have a proper namespace + static constexpr variables.

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
+#ifndef MPT_UTILITY_H
+#define MPT_UTILITY_H
```
mathbunnyru, line 2: pragma once

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+
+#include <cstddef>
+#include <secp256k1.h>
+#include <stdint.h>
```
mathbunnyru, line 7:
```suggestion
#include <cstdint>
```

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+    (kMPT_COMMON_HASH_SIZE + kMPT_ACCOUNT_ID_SIZE + kMPT_VERSION_SIZE)  // 74 bytes
+#define kMPT_CONVERT_BACK_HASH_SIZE \
+    (kMPT_COMMON_HASH_SIZE + kMPT_AMOUNT_SIZE + kMPT_VERSION_SIZE)  // 62 bytes
+#define kMPT_CLAWBACK_HASH_SIZE \
+    (kMPT_COMMON_HASH_SIZE + kMPT_AMOUNT_SIZE + kMPT_ACCOUNT_ID_SIZE)  // 78 bytes
+
+/**
+ * @brief Represents a unique 24-byte MPT issuance ID.
+ */
+typedef struct
```
mathbunnyru, line 50: I don't think we need `typedef struct` in C++ and can simply declare struct's (and below as well)

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+{
+    uint8_t bytes[kMPT_ISSUANCE_ID_SIZE];
+} mpt_issuance_id;
```
mathbunnyru, line 53: I don't think it's the codestyle we use for structs/classes (and below as well)

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+/**
+ * @brief Context Hash for ConfidentialMPTConvert.
+ */
+int
+mpt_get_convert_context_hash(
+    account_id account,
+    uint32_t sequence,
+    mpt_issuance_id issuanceID,
+    uint64_t amount,
+    uint8_t out_hash[kMPT_HALF_SHA_SIZE]);
```
mathbunnyru, line 107: In C++, we don't take out params as input (and many params below as well)

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+#define MPT_UTILITY_H
+
+#include <cstddef>
+#include <secp256k1.h>
+#include <stdint.h>
+#include <vector>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
mathbunnyru, line 11: Doesn't make sense (and in the end of file), since you include vector

tests/test_mpt_utility.cpp
```diff
@@ -0,0 +1,371 @@
...
+    }
+}
+
+void
+test_mpt_confidential_convert()
+{
+    account_id acc;
+    std::memset(acc.bytes, 0xAA, kMPT_ACCOUNT_ID_SIZE);
+    mpt_issuance_id issuance;
+    std::memset(issuance.bytes, 0xBB, kMPT_ISSUANCE_ID_SIZE);
```
mathbunnyru: In C++ we have ctors for this

mathbunnyru (changes requested):
It seems that `mpt_utility.h` is a C++ only header, but some things are implemented in C, some in C++.
I would prefer sticking with one and not mixing them.

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+
+    /**
+     * @brief The 32-byte secret random value used to blind the Pedersen commitment.
+     */
+    uint8_t blinding_factor[kMPT_BLINDING_FACTOR_SIZE];
+};
+
+secp256k1_context*
+mpt_secp256k1_context();
+
```
mrtcnk, line 98: Small note: can we ensure the TransactionContextID construction here matches the spec 
H(TxType ∥ Account ∥ MPTokenIssuanceID ∥ SequenceOrTicket ∥ TxSpecific)
where:
- Send / ConvertBack: TxSpecific = Receiver ∥ CBSVersion
- Convert: TxSpecific = Account ∥ 0

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+// General crypto primitive sizes in bytes
+#define kMPT_HALF_SHA_SIZE 32
+#define kMPT_PUBKEY_SIZE 64
+#define kMPT_PRIVKEY_SIZE 32
+#define kMPT_BLINDING_FACTOR_SIZE 32
+
+// Gamal & Pedersen primitive sizes in bytes
+#define kMPT_GAMAL_CIPHER_SIZE 33
+#define kMPT_GAMAL_TOTAL_SIZE 66
```
mrtcnk: Nit: should we prefer ELGAMAL instead of GAMAL in the constant names?

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+#define ttCONFIDENTIAL_MPT_MERGE_INBOX 86
+#define ttCONFIDENTIAL_MPT_CONVERT_BACK 87
+#define ttCONFIDENTIAL_MPT_SEND 88
+#define ttCONFIDENTIAL_MPT_CLAWBACK 89
+
+// General crypto primitive sizes in bytes
+#define kMPT_HALF_SHA_SIZE 32
+#define kMPT_PUBKEY_SIZE 64
+#define kMPT_PRIVKEY_SIZE 32
+#define kMPT_BLINDING_FACTOR_SIZE 32
```
mrtcnk, line 24: Quick check: Can we confirm the sizes here, particularly for public key encoding (compressed 33-byte vs uncompressed 64-byte)?

mrtcnk (commented):
Left a few comments mostly as reference points. Nothing blocking from my side. Happy to iterate further in follow-up PRs as the utility layer evolves.

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,381 @@
...
+    uint8_t out[kMPT_GAMAL_TOTAL_SIZE]);
+
+/**
+ * @brief Generates a new Secp256k1 ElGamal keypair.
+ * @param out_privkey [out] A 32-byte buffer for private key.
+ * @param out_pubkey  [out] A 64-byte buffer for public key.
+ * @return 0 on success, -1 on failure.
+ */
+int
+mpt_generate_keypair(uint8_t* out_priv, uint8_t* out_pub);
```
Patel-Raj11: @yinyiqian1 Hi, Quick question, since we are not exposing `generate_canonical_encrypted_zero`, I assume that we won't need it on client side for proof generation. Do we use it for proof verification only?

yinyiqian1: `generate_canonical_encrypted_zero` is not required for ZK proof generation or verification. It is specifically used for the internal representation of an empty confidential balance on the ledger. The logic is encapsulated in `rippled`'s transactions. Client does not need to use it explicitly.

Patel-Raj11: Thanks for clarifying.

yinyiqian1: You are welcome.

include/utility/mpt_utility.h
yinyiqian1: Yes. We just modified rippled from 64 uncompressed to 33-byte compressed key. This will be adapted as well. There will be a new commit adapting this.

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,642 @@
...
+    return 0;
+}
+
+int
+mpt_generate_blinding_factor(uint8_t out_factor[kMPT_BLINDING_FACTOR_SIZE])
+{
+    if (!out_factor)
+        return -1;
+
+    if (RAND_bytes(out_factor, kMPT_BLINDING_FACTOR_SIZE) != 1)
```
Patel-Raj11: @yinyiqian1 Do we need to check if blinding factor is valid using `secp256k1_ec_seckey_verify`?
In `test_encryption` unit test, we are using `secp256k1_elgamal_generate_keypair` to generate blinding factor so, I am little confused here.

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,642 @@
...
+        offset += 8;
+    }
+
+    void
+    addRaw(uint8_t const* data, size_t len)
+    {
+        memcpy(buffer + offset, data, len);
+        offset += len;
+    }
+};
```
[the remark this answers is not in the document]
yinyiqian1: https://github.com/XRPLF/mpt-crypto/pull/12/changes/5fc63043a6ff3a51a8b374e3f3b269d70dfde79a

include/utility/mpt_utility.h
yinyiqian1: updated in: https://github.com/XRPLF/mpt-crypto/pull/12/changes/5fc63043a6ff3a51a8b374e3f3b269d70dfde79a

include/utility/mpt_utility.h
yinyiqian1: Adapted to compressed key: https://github.com/XRPLF/mpt-crypto/pull/12/changes/768a3630ad8c0e25e1a5b4250dbea766b3705c94

src/utility/mpt_utility.cpp
yinyiqian1: Good catch! I just updated here: https://github.com/XRPLF/mpt-crypto/commit/f2f5a0b45851547f00ac761f7892682804eb02b8
Thank you for the finding.

Patel-Raj11: @yinyiqian1 I didn't meant to request a change here. But, good to know that blinding factor or any random scalar (generated using `generate_random_scalar`) needs to be verified using `secp256k1_ec_seckey_verify`.

include/utility/mpt_utility.h
yinyiqian1: I think I should keep `stdint.h`, and change <cstddef> to <stdint.h> for C/C++ compatibility:

```
#include <secp256k1.h>
#include <stddef.h>
#include <stdint.h>
```

include/utility/mpt_utility.h
yinyiqian1: I think I should move `#include <vector>` to `.cpp`, and keep the header in C style. https://github.com/XRPLF/mpt-crypto/pull/12/changes/65eaa4c89bd8f373b234e171f30c80eb57820c64

include/utility/mpt_utility.h
yinyiqian1: I'll keep #ifndef to match this lib's style and portability.

include/utility/mpt_utility.h
yinyiqian1: I'd prefer to keep the header in pure C style to ensure maximum portability for non-C++ callers

include/utility/mpt_utility.h
yinyiqian1: I'd prefer to keep the header in pure C style to ensure maximum portability for non-C++ callers

include/utility/mpt_utility.h
yinyiqian1: the header is exposing as C style for portability

include/utility/mpt_utility.h
yinyiqian1: moved vector to .cpp, the header now only maintains C style

tests/test_mpt_utility.cpp
yinyiqian1: I want to keep the header file in C style with POD.
So I'll create a helper in the test cpp file, the usage of this helper is limited to this test file:

```
// helper to create mock accounts and issuance IDs
template <typename T>
T create_mock_id(uint8_t fill) {
    T mock;
    std::fill(std::begin(mock.bytes), std::end(mock.bytes), fill);
    return mock;
}
```
I also modified the C style memcpy to std::copy in this commit: https://github.com/XRPLF/mpt-crypto/pull/12/changes/902432720464326da27fe2b0a0fabfcb826099ee

tests/test_mpt_utility.cpp
yinyiqian1: https://github.com/yinyiqian1/mpt-crypto/blob/902432720464326da27fe2b0a0fabfcb826099ee/tests/test_mpt_utility.cpp#L28
I added more details in the comment. I'll remove it once our lib can deal with these large numbers.

include/utility/mpt_utility.h
yinyiqian1: The context hash is adapting to the rippled's implementation hash. We'll change on both sides if anything requires modification.

The current hashing is like: 
We have a `common hash` matches `TxType ∥ Account ∥ MPTokenIssuanceID ∥ SequenceOrTicket`, see the function `mpt_add_common_zkp_fields`.

`Convert`: `TxSpecific = amount`. (see `mpt_get_convert_context_hash`). because we are trying to add all the inputs into the hash function. (account already included in the common hash)
`ConvertBack `: `TxSpecific = amount | CBSVersion`.    (account already included in the common hash)
`Send`: TxSpecific = `receiver | CBSVersion`.
`Clawback`: `TxSpecific = amount | holder`

The difference with the doc: 
- introduced for convert/convertback/clawback.
- some TxSpecific fields omit the Account field, since it is already included in the common hash section.

What we will add according to what you suggested in the doc:
- We need to pass in the ticket number if it is a ticket, although the interface won't change.
- We'll add version 0 for convert: `TxSpecific = amount | 0`
 
Please suggest if this looks good to you or any other modifications needed?

mrtcnk: Thanks for the detailed breakdown! The common hash structure aligns well with the spec. From the spec side, I’d prefer that we keep the TransactionContextID definition. The cryptographic intent there is that TxSpecific binds the transaction context and state freshness (identity||version), rather than public payload inputs like the plaintext amount. For example, in ConfidentialMPTConvert, the spec intentionally defines TxSpecific := Account||0 to maintain a uniform semantic structure across transaction types, where 0 explicitly indicates that no CBSVersion is involved. The amount is already bound inside the Fiat–Shamir transcript (mG) during challenge computation, so it does not need to be part of TransactionContextID.

To keep the transcript definition consistent with the spec, my preference for the semantic shape of TxSpecific would be:

ConfidentialMPTSend: receiver||CBSVersion
ConfidentialMPTConvertBack: receiver (=Account)||CBSVersion
ConfidentialMPTConvert: account||0
ConfidentialMPTClawback: holder||0

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,642 @@
+#include <arpa/inet.h>
```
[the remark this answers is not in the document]
pdp2121: Hi @yinyiqian1 , can this issue be fixed? I had to create a patch on our end to handle Windows. Probably something like this should work:
```
#ifdef _WIN32
    #include <winsock2.h>
#else
    #include <arpa/inet.h>
#endif
```

include/utility/mpt_utility.h
```diff
@@ -0,0 +1,384 @@
...
+typedef struct
+{
+    uint8_t bytes[kMPT_ACCOUNT_ID_SIZE];
+} account_id;
+
+/**
+ * @brief Represents a recipient in a Confidential Send transaction.
+ * Contains a pubkey and an encrypted amount.
+ */
+struct mpt_confidential_recipient
```
Patel-Raj11: @yinyiqian1 Hi, Would it make sense to call this struct as `mpt_confidential_participant` or something else that does not imply that it is a recipient of the confidential amount? As other then the receiver all other are the participants of the send transaction?

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,775 @@
...
+            return -1;
+
+        if (!secp256k1_ec_pubkey_parse(
+                ctx, &s[i], rec.encrypted_amount + kMPT_ELGAMAL_CIPHER_SIZE, kMPT_ELGAMAL_CIPHER_SIZE))
+            return -1;
+
+        if (secp256k1_ec_pubkey_parse(ctx, &pk[i], rec.pubkey, kMPT_PUBKEY_SIZE) != 1)
+            return -1;
+
+        sr.insert(sr.end(), tx_blinding_factor, tx_blinding_factor + kMPT_BLINDING_FACTOR_SIZE);
```
Patel-Raj11: I think if each participant/recipient has different blinding factor, the proof generation and verification both works. Since there is no BlindingFactor field for ConfidentialMPTSend transaction type, different blinding factor's can be used by each participant to encrypt the same amount.
Could you confirm this on your end once and if it's true then should we append multiple blinding factors here?

yinyiqian1: No. We should use the same blinding factor for encrypting the participant/recipient. The blinding factor is generated by the sender off-chain. It is a value that zkproof needs to verify, it should not be part of the request.
If we use different blinding factors for the ElGamal ciphertexts, the Equality Proof inside the ZK-Proof would fail.

We can use another set of blinding factor for pedersen commitment. But for the ElGamal ciphertexts, we must use the same blinding factor for issuer,auditor and sender.


Btw, in ConfidentialMPTConvert, it is a public amount, we can put the blinding factor in the request. We are not trying to prove the knowledge of the randomness in that case. We are proving the knowledge of private key in that case, using schnorr proof. 

Key difference:
- convert: prove the knowledge of identity (private key)
- send (equality proof part):  prove the knowledge of randomness (blinding factor) and the actual value we are encrypting.

Patel-Raj11: > If we use different blinding factors for the ElGamal ciphertexts, the Equality Proof inside the ZK-Proof would fail.

I have an integration test that runs on this [commit](https://github.com/XRPLF/rippled/pull/6404) and that test uses different blinding factors for sender, receiver and issuer. The proof that gets submitted to rippled still gets verified successfully. `test_valid_multi_proof` in mpt-crypto also verifies with different blinding factor. So wanted to know if something changed recently or it's some logic in rippled verification part that requires same blinding factor to be used for encryption for Send transaction?

yinyiqian1: @Patel-Raj11 Sorry for the confusion. We're going to have an optimization that is forcing the shared r. This was newly supported from the `proof_same_plaintext_multi_shared_r.c`. By that time, different blinding factor will fail.

include/utility/mpt_utility.h
yinyiqian1: @Patel-Raj11 Thank you. Yes, it makes sense. How about `mpt_confidential_party`, which is shorter, but similar to participant?

src/utility/mpt_utility.cpp
Patel-Raj11: Ok, sounds good. I'll change my implementation now to use same blinding factor, but there will be change in the utility functions to use `secp256k1_mpt_prove_equality_shared_r` right?

include/utility/mpt_utility.h
Patel-Raj11: `mpt_confidential_party` sounds good to me.

src/utility/mpt_utility.cpp
yinyiqian1: Yes. That is correct

yinyiqian1: But I'll update the Rippled's c++ code first, and then make the changes in this SDK lib.

include/utility/mpt_utility.h
yinyiqian1: @mrtcnk Just want to confirm, do we use `ConfidentialMPTClawback: holder||0` on purpose?  Or should we use `holder || version` instead ?  I feel that the balance of the holder changed, and his version is updated, so we might be better use his version instead of 0.

so in the paper 3.3.1, 
```
Send and ConvertBack binding. For transaction types whose soundness depends on a specific snapshot of
the sender’s confidential spending balance—namely ConfidentialMPTSend and ConfidentialMPTConvertBack—we define
TxSpecific := Receiver ∥ CB_S_Version.
```
If we want to use the holder's version, maybe we can add ConfidentialMPTClawback in this paragraph as well.

src/utility/mpt_utility.cpp
yinyiqian1: updated https://github.com/XRPLF/mpt-crypto/pull/12/changes/91a04136e821791c6b1f4b8b43651a6d64707492

src/utility/mpt_utility.cpp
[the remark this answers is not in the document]
yinyiqian1: just updated in https://github.com/XRPLF/mpt-crypto/pull/12/changes/7f68120ea2355d50ac5b217b2926839a1b2b78fc

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,642 @@
...
+    size_t* out_len)
+{
+    if (!priv || !recipients || !out_proof || !out_len)
+        return -1;
+
+    secp256k1_context const* ctx = mpt_secp256k1_context();
+    if (!ctx)
+        return -1;
+
+    std::vector<secp256k1_pubkey> r(n_recipients);
```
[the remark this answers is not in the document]
yinyiqian1: updated with:
```
    if (n_recipients != 3 && n_recipients != 4)
      return -1;
```

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,1262 @@
...
+            ctx,
+            out,
+            &pk,
+            &c2,
+            &c1,
+            &pcm,
+            params->amount,
+            priv,
+            params->blinding_factor,
+            context_hash) != 1)
```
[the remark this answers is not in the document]
yinyiqian1: For balance Pedersen linkage, we intended to use this order, which is counterintuitive.
the call order of c1 c2 pk is referring to the amount Pedersen linkage.

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,1262 @@
...
+    secp256k1_pubkey pk;
+    if (secp256k1_ec_pubkey_parse(ctx, &pk, pub, kMPT_PUBKEY_SIZE) != 1)
+        return -1;
+
+    secp256k1_pubkey c1, c2;
+    if (!mpt_make_ec_pair(ciphertext, c1, c2))
+        return -1;
+
+    if (secp256k1_equality_plaintext_prove(
+            ctx, out_proof, &pk, &c2, &c1, amount, priv, context_hash) != 1)
```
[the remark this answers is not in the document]
yinyiqian1: This is by design. ` &pk, &c2, &c1` is the correct order.

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,1262 @@
...
+
+    secp256k1_pubkey c1, c2;
+    if (!mpt_make_ec_pair(ciphertext, c1, c2))
+        return -1;
+
+    secp256k1_pubkey pk;
+    if (secp256k1_ec_pubkey_parse(ctx, &pk, pubkey, kMPT_PUBKEY_SIZE) != 1)
+        return -1;
+
+    if (secp256k1_equality_plaintext_verify(ctx, proof, &pk, &c2, &c1, amount, context_hash) != 1)
```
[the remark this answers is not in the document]
yinyiqian1: This is by design.  &pk, &c2, &c1 is the correct order for clawback

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,1262 @@
...
+    if (!mpt_make_ec_pair(ciphertext, c1, c2))
+        return -1;
+
+    if (secp256k1_ec_pubkey_parse(ctx, &pk, pubkey, kMPT_PUBKEY_SIZE) != 1)
+        return -1;
+
+    if (secp256k1_ec_pubkey_parse(ctx, &pcm, commitment, kMPT_PEDERSEN_COMMIT_SIZE) != 1)
+        return -1;
+
+    if (secp256k1_elgamal_pedersen_link_verify(ctx, proof, &pk, &c2, &c1, &pcm, context_hash) != 1)
```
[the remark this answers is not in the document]
yinyiqian1: this order is by design

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,1262 @@
...
+    for (int i = 0; i < 8; ++i)
+    {
+        scalar[31 - i] = static_cast<uint8_t>(amount >> (i * 8));
+    }
+
+    // Calculate mG and negate it to get -mG
+    secp256k1_pubkey mG;
+    if (secp256k1_ec_pubkey_create(ctx, &mG, scalar) != 1)
+        return -1;
+
```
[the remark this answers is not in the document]
yinyiqian1: added fast path in the code:
```
    // Subtracting zero leaves the commitment unchanged
    if (amount == 0)
    {
        std::memcpy(remainder, commitment_in, kMPT_PEDERSEN_COMMIT_SIZE);
        return 0;
    }
```

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,1262 @@
...
+    if (!ctx)
+        return -1;
+
+    // todo: zero amount
+    if (amount == 0)
+    {
+        std::memset(out_commitment, 0, kMPT_PEDERSEN_COMMIT_SIZE);
+        return 0;
+    }
+
```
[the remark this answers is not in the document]
yinyiqian1: removed 0 handling

src/utility/mpt_utility.cpp
```diff
@@ -0,0 +1,642 @@
...
```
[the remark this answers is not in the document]
yinyiqian1: removed

mathbunnyru (approved):
I haven't verified the code itself, but as soon as @mrtcnk approves, I'm ready to merge

include/utility/mpt_utility.h
mrtcnk: Clawback is issuer-initiated and requires issuer authorization, so using the holder’s CBS_version here may introduce an unnecessary dependency on holder-side state. So, we can keep using holder || 0.

mrtcnk (approved):
Looks good to me. Thanks for addressing the review comments.

PR #13, opened 2026-02-19
fix: resolve zero-value proof failure in tests

mrtcnk:
### Description

When a user attempts to prove `value = 0`, the decomposed left bit vector ($\mathbf{a}_L$) is entirely zeros. Similarly, if proving the maximum value (`UINT64_MAX`), the right bit vector ($\mathbf{a}_R$) becomes entirely zeros. When these all-zero vectors are passed to the internal Multi-Scalar Multiplication (`ipa_msm`) function, the mathematical result is the Point at Infinity. Since the `libsecp256k1` library cannot represent infinity in standard affine coordinates, the MSM function correctly returns `0`. Previously, the prover treated this `0` return value as a fatal library error and aborted the proof generation.

**Fix:**
Updated `secp256k1_bulletproof_prove_agg` to handle the `0` return from `ipa_msm`. If the result evaluates to the Point at Infinity, the prover now safely skips adding that empty term to the commitment instead of failing. There are no changes to the underlying math library or external APIs.

### Changes
- Modified the point combination logic for Commitments `A` and `S` in `src/bulletproof_aggregated.c` to safely handle the Point at Infinity.
- Added comprehensive explicit test cases to `tests/test_bulletproof_agg.c`:
  - `m=1, value={0}` (The exact bug reproduction).
  - `m=1, value={UINT64_MAX}` (Tests the opposite boundary condition where the right-side bit vector becomes all zeros).
  - `m=2, values={0, 0}` (Ensures zero-term skipping works correctly across aggregated proofs).
  - `m=1, value={1}` and `m=2, values={0, 1}` (Ensures standard non-zero proofs still pass without regression).

### DIFF
--- a/src/bulletproof_aggregated.c
+++ b/src/bulletproof_aggregated.c
@@ -1220,20 +1220,49 @@ int secp256k1_bulletproof_prove_agg(
     {
         secp256k1_pubkey tG, tH, tB;
         const secp256k1_pubkey* pts[3];
+        int n_pts = 0;
 
-        if (!secp256k1_bulletproof_ipa_msm(ctx, &tG, G_vec, al, n)) goto cleanup;
-        if (!secp256k1_bulletproof_ipa_msm(ctx, &tH, H_vec, ar, n)) goto cleanup;
+        /* A Calculation */
+        /* 1. alpha * Base (Always exists) */
         tB = *pk_base;
         if (!secp256k1_ec_pubkey_tweak_mul(ctx, &tB, alpha)) goto cleanup;
-        pts[0] = &tB; pts[1] = &tG; pts[2] = &tH;
-        if (!secp256k1_ec_pubkey_combine(ctx, &A, pts, 3)) goto cleanup;
+        pts[n_pts++] = &tB;
 
-        if (!secp256k1_bulletproof_ipa_msm(ctx, &tG, G_vec, sl, n)) goto cleanup;
-        if (!secp256k1_bulletproof_ipa_msm(ctx, &tH, H_vec, sr, n)) goto cleanup;
+        /* 2. <al, G> (Only add if al is non-zero) */
+        /* Note: ipa_msm returns 0 if result is infinity (all scalars 0). This is valid for value=0. */
+        if (secp256k1_bulletproof_ipa_msm(ctx, &tG, G_vec, al, n)) {
+            pts[n_pts++] = &tG;
+        }
+
+        /* 3. <ar, H> (Only add if ar is non-zero) */
+        if (secp256k1_bulletproof_ipa_msm(ctx, &tH, H_vec, ar, n)) {
+            pts[n_pts++] = &tH;
+        }
+
+        /* Combine valid points for A */
+        if (!secp256k1_ec_pubkey_combine(ctx, &A, pts, n_pts)) goto cleanup;
+
+
+        /* S Calculation (Follows same logic for consistency) */
+        n_pts = 0;
+
+        /* 1. rho * Base */
         tB = *pk_base;
         if (!secp256k1_ec_pubkey_tweak_mul(ctx, &tB, rho)) goto cleanup;
-        pts[0] = &tB; pts[1] = &tG; pts[2] = &tH;
-        if (!secp256k1_ec_pubkey_combine(ctx, &S, pts, 3)) goto cleanup;
+        pts[n_pts++] = &tB;
+
+        /* 2. <sl, G> */
+        if (secp256k1_bulletproof_ipa_msm(ctx, &tG, G_vec, sl, n)) {
+            pts[n_pts++] = &tG;
+        }
+
+        /* 3. <sr, H> */
+        if (secp256k1_bulletproof_ipa_msm(ctx, &tH, H_vec, sr, n)) {
+            pts[n_pts++] = &tH;
+        }
+
+        /* Combine valid points for S */
+        if (!secp256k1_ec_pubkey_combine(ctx, &S, pts, n_pts)) goto cleanup;
     }
 
     /* ---- 6. Fiat–Shamir y,z ---- */
--- a/src/bulletproof_aggregated.c
+++ b/src/bulletproof_aggregated.c
@@ -1122,7 +1122,7 @@ static int calculate_commitment_term(
     if (!secp256k1_ec_pubkey_tweak_mul(ctx, &tB, base_scalar)) return 0;
     pts[n_pts++] = &tB;
 
-   /* 2. <vec_l, G> */
+    /* 2. <vec_l, G> */
     if (!scalar_vector_all_zero(vec_l, n)) {
         if (!secp256k1_bulletproof_ipa_msm(ctx, &tG, G_vec, vec_l, n))
             return 0; /* REAL FAILURE */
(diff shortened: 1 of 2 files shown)


src/bulletproof_aggregated.c
```diff
@@ -1220,20 +1220,49 @@ int secp256k1_bulletproof_prove_agg(
...
+        /* 3. <ar, H> (Only add if ar is non-zero) */
+        if (secp256k1_bulletproof_ipa_msm(ctx, &tH, H_vec, ar, n)) {
+            pts[n_pts++] = &tH;
+        }
+
+        /* Combine valid points for A */
+        if (!secp256k1_ec_pubkey_combine(ctx, &A, pts, n_pts)) goto cleanup;
+
+
+        /* S Calculation (Follows same logic for consistency) */
```
mathbunnyru: Can we maybe have a function and call it twice, if the logic is the same?

tests/test_bulletproof_agg.c
```diff
@@ -6,184 +6,134 @@
...
 static inline double elapsed_ms(struct timespec a, struct timespec b) {
     return (b.tv_sec - a.tv_sec) * 1000.0 +
            (b.tv_nsec - a.tv_nsec) / 1e6;
 }
 
-/* ---- Main ---- */
-int main(void) {
-    printf("[TEST] Aggregated Bulletproof test (m = %d)\n", M);
+/* ---- Core Test Logic ---- */
+void run_test_case(secp256k1_context* ctx, const char* name, uint64_t* values, size_t m, int run_benchmarks) {
```
mathbunnyru: `m` is not a clear name variable and usually is bad practice (except maybe short functions/loops)
Could you please name it better?

tests/test_bulletproof_agg.c
```diff
@@ -6,184 +6,134 @@
...
-    for (size_t i = 0; i < M; i++) {
+    for (size_t i = 0; i < m; i++) {
         random_scalar(ctx, blindings[i]);
         EXPECT(secp256k1_bulletproof_create_commitment(
-                ctx,
-                &commitments[i],
-                values[i],
-                blindings[i],
-                &pk_base));
+                ctx, &commitments[i], values[i], blindings[i], &pk_base));
```
mathbunnyru: I think this is purely style change. Please avoid making style changes in feature/fix PRs.
This comment applies to some of changes below as well

mathbunnyru (changes requested):
I've made some small comments about code quality / PR structure.
Please, ask someone with domain knowledge to review the code itself.

src/bulletproof_aggregated.c
mrtcnk: Thanks for the proposal! I've extracted the common logic into a static helper function calculate_commitment_term.

tests/test_bulletproof_agg.c
mrtcnk: Makes total sense. I've updated the variable name to num_values throughout the test function for better clarity.

tests/test_bulletproof_agg.c
mrtcnk: Thanks for this! I’ve reverted the formatting of those function calls back to their original style to keep the diff clean.

tests/test_bulletproof_agg.c
```diff
@@ -6,46 +6,30 @@
...
-
-    /* ---- Values ---- */
-    uint64_t values[M] = { 5000, 123456 };
-    unsigned char blindings[M][32];
-    secp256k1_pubkey commitments[M];
-
-    /* ---- Context Binding ---- */
+    unsigned char blindings[num_values][32];
+    secp256k1_pubkey commitments[num_values];
     unsigned char context_id[32];
```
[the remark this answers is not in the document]
mrtcnk: Done. I've removed the variable-length arrays

mathbunnyru (dismissed):
LGTM, please ask someone with domain knowledge to review the code itself.

src/bulletproof_aggregated.c
```diff
@@ -1092,6 +1092,40 @@ int secp256k1_bulletproof_create_commitment(
...
+
+    /* 2. <vec_l, G> */
+    if (secp256k1_bulletproof_ipa_msm(ctx, &tG, G_vec, vec_l, n)) {
+        pts[n_pts++] = &tG;
+    }
+
+    /* 3. <vec_r, H> */
+    if (secp256k1_bulletproof_ipa_msm(ctx, &tH, H_vec, vec_r, n)) {
+        pts[n_pts++] = &tH;
+    }
```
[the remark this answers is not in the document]
mrtcnk: Updated calculate_commitment_term() to explicitly distinguish the point-at-infinity case from real MSM failures.

tests/test_bulletproof_agg.c
```diff
@@ -105,59 +84,51 @@ int main(void) {
...
     random_scalar(ctx, bad_blinding);
 
-    /* Create a fake commitment to (value + 1) to break the sum */
+    /* Create fake commitment to (value + 1) */
     EXPECT(secp256k1_bulletproof_create_commitment(
             ctx,
-            &bad_commitments[1],
-            values[M - 1] + 1,
+            &bad_commitments[num_values - 1],
+            values[num_values - 1] + 1,
```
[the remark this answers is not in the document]
mrtcnk: Added EXPECT(num_values > 0) at the start of run_test_case() to make the implicit assumption explicit.

src/bulletproof_aggregated.c
```diff
@@ -1091,6 +1091,53 @@ int secp256k1_bulletproof_create_commitment(
 
     return 1;
 }
+/* Helper for a vector 0 */
```
mathbunnyru, line 1095:
```suggestion
}

/* Helper for a vector 0 */
```

src/bulletproof_aggregated.c
```diff
@@ -1091,6 +1091,53 @@ int secp256k1_bulletproof_create_commitment(
...
+    secp256k1_pubkey tG, tH, tB;
+    const secp256k1_pubkey* pts[3];
+    int n_pts = 0;
+
+    /* 1. base_scalar * Base */
+    tB = *pk_base;
+    if (!secp256k1_ec_pubkey_tweak_mul(ctx, &tB, base_scalar)) return 0;
+    pts[n_pts++] = &tB;
+
+   /* 2. <vec_l, G> */
```
mathbunnyru:
```suggestion
    /* 2. <vec_l, G> */
```

src/bulletproof_aggregated.c
```diff
@@ -1091,6 +1091,53 @@ int secp256k1_bulletproof_create_commitment(
...
+        pts[n_pts++] = &tG;
+    }
+
+    /* 3. <vec_r, H> */
+    if (!scalar_vector_all_zero(vec_r, n)) {
+        if (!secp256k1_bulletproof_ipa_msm(ctx, &tH, H_vec, vec_r, n))
+            return 0; /* REAL FAILURE */
+        pts[n_pts++] = &tH;
+    }
+
```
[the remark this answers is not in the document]
mrtcnk: Added an explicit fast-path when n_pts == 1.

tests/test_bulletproof_agg.c
```diff
@@ -105,59 +85,51 @@ int main(void) {
...
```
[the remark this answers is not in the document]
mrtcnk: Updated the test for the MAX value case.

mathbunnyru (approved):

PR #22, opened 2026-04-07
Migrate utility layer to compact sigma proofs

tesseract-ripple:
## Summary

- Replace legacy commitment-form proof calls in `mpt_utility.cpp` with the compact sigma prove/verify functions introduced in PR #18
- Remove 3 legacy proof implementations (`equality_proof.c`, `proof_link.c`, `proof_same_plaintext_multi_shared_r.c`) and their tests (~1800 lines deleted)
- Update proof-size constants in `mpt_utility.h`

**Breaking change:** This changes the wire format of all sigma proofs consumed by rippled. The proof sizes and serialization formats differ from the commitment-form proofs on `main`. Rippled's `ConfidentialTransfer` code must be updated to use the new proof-size constants and the change must be gated behind an XRPL amendment, since validators must agree on the accepted proof format.

## Proof size changes

| Transaction | Old sigma | New sigma | Saving |
|-------------|-----------|-----------|--------|
| Send (n=4) | 619B | 192B | 427B (69%) |
| ConvertBack | 195B | 128B | 67B (34%) |
| Clawback | 98B | 64B | 34B (35%) |

Bulletproof range proofs are unchanged.

## Migration mapping

| Utility function | Old calls | New call |
|-----------------|-----------|----------|
| `mpt_get_send_proof` | `prove_equality_shared_r` + 2x `link_prove` | `secp256k1_compact_standard_prove` |
| `mpt_get_clawback_proof` | `equality_plaintext_prove` | `secp256k1_compact_clawback_prove` |
| `mpt_get_convert_back_proof` | `link_prove` | `secp256k1_compact_convertback_prove` |

Verify functions follow the same pattern.

## Test plan

- [x] 10/10 tests pass (3 legacy tests removed, 10 remaining)
- [ ] Verify `test_mpt_utility` covers send, clawback, and convertback round-trips through the new compact paths
- [ ] Review that removed files have no remaining callers
- [ ] Coordinate rippled-side proof-size constant updates and amendment gating

PR #24, opened 2026-04-10
Integrate Compact proof into mpt_utility

yinyiqian1:
Replaces the legacy equality and linkage proofs across all Confidential MPT transaction types with compact AND-composed sigma proofs:

- `ConfidentialMPTConvert`: 64 byte.
- `ConfidentialMPTSend`: 192 byte compact sigma + 754 byte bullet proof = 946 byte.
- `ConfidentialMPTConvertBack`: 128 byte compact sigma + 688 byte bullet proof = 816 byte
- `ConfidentialMPTClawback`: 64 byte.


mrtcnk (approved):
Everything looks good to me. The code is clean, the tests are passing, and the redundant functions have been properly removed.

tesseract-ripple (commented):
(Review superseded — PR already merged.)

include/utility/mpt_utility.h
```diff
@@ -414,27 +383,24 @@ mpt_verify_convert_back_proof(
...
+ *                                        participants[0] is the sender.
+ * @param n_participants             [in] Number of participants (3 or 4).
+ * @param sender_spending_ciphertext [in] The sender's on-ledger balance ciphertext (b1||b2).
+ * @param amount_commitment          [in] Pedersen commitment pc_m to the transfer amount.
+ * @param balance_commitment         [in] Pedersen commitment pc_b to the sender's balance.
  * @param context_hash               [in] The 32-byte transaction context hash.
  * @return 0 on success, -1 on failure.
  */
 int
 mpt_verify_send_proof(
```
tesseract-ripple, line 402: **Q: Dropping `proof_len` from the verifier — intentional?**

The old `mpt_verify_send_proof` accepted a `proof_len` parameter and validated it against the expected total before reading the buffer. This version drops it — the verifier reads a fixed 946 bytes with no length check.

Was this a deliberate simplification (fixed-size proof means the caller should know the size), or an oversight? Our parallel branch (PR #22) kept `proof_len` as defense-in-depth — a caller passing a short buffer gets `-1` instead of UB. Curious whether rippled's call site already guarantees the buffer size through other means.

include/utility/mpt_utility.h
```diff
@@ -257,77 +250,64 @@ mpt_get_pedersen_commitment(
...
+ * @param context_hash       [in] The 32-byte context hash.
+ * @param amount_commitment  [in] Pedersen commitment pc_m = m*G + r*H.
+ * @param balance_params     [in] Includes pedersen_commitment (pc_b), amount (balance),
+ *                                blinding_factor (rho), and ciphertext (b1||b2).
+ * @param out_proof          [out] Buffer to receive the proof blob.
+ * @param out_len            [in/out] In: capacity (must be >= 946). Out: bytes written.
+ * @return 0 on success, -1 on failure.
  */
 int
 mpt_get_confidential_send_proof(
```
tesseract-ripple, line 281: **Q: Adding `pub` parameter + raw `amount_commitment` — what drove this?**

Two changes here vs. the old API:
1. The sender's `pub` key is now passed explicitly, where before it was derived internally from `priv` via `secp256k1_ec_pubkey_create`.
2. `amount_params` (a `mpt_pedersen_proof_params*`) is replaced with raw `amount_commitment` bytes.

For (1): is this because rippled already has the pubkey available and you wanted to avoid redundant derivation? Or is there a correctness reason (e.g., the pubkey on the ledger might not match `priv * G` in some edge case)?

For (2): the old `amount_params` struct bundled the commitment with its blinding factor and amount. Now the prover takes `amount_commitment` as raw bytes but still needs `tx_blinding_factor` and `amount` separately. Is the intent to simplify the caller's interface, or was the struct causing issues on the rippled side?

src/utility/mpt_utility.cpp
```diff
@@ -767,65 +699,15 @@ mpt_get_clawback_proof(
...
 
-    if (secp256k1_equality_plaintext_prove(
-            ctx, out_proof, &pk, &c2, &c1, amount, priv, context_hash) != 1)
+    if (secp256k1_compact_clawback_prove(
+            ctx, out_proof, amount, priv, &pk, &c1, &c2, context_hash) != 1)
     {
         return -1;
     }
 
     return 0;
```
tesseract-ripple, line 708: **Q: Removing `mpt_verify_revealed_amount` — still used in rippled?**

This function verified ElGamal encryption correctness without a ZKP (direct blinding-factor reveal). It was used in `ConfidentialMPTConvert` on the rippled side. Is it no longer needed, or does rippled have its own implementation now?

include/utility/mpt_utility.h
```diff
@@ -336,17 +316,17 @@ mpt_get_convert_back_proof(
     uint8_t const context_hash[kMPT_HALF_SHA_SIZE],
     uint64_t const amount,
     mpt_pedersen_proof_params const* params,
-    uint8_t out_proof[kMPT_PEDERSEN_LINK_SIZE + kMPT_SINGLE_BULLETPROOF_SIZE]);
+    uint8_t out_proof[SECP256K1_COMPACT_CONVERTBACK_PROOF_SIZE + kMPT_SINGLE_BULLETPROOF_SIZE]);
```
tesseract-ripple, line 319: Nit: the proof-size constants here use `SECP256K1_COMPACT_CONVERTBACK_PROOF_SIZE` (from `secp256k1_mpt.h`) directly in the utility header. This means `mpt_utility.h` now depends on `secp256k1_mpt.h` for these macros (and indeed adds `#include <secp256k1_mpt.h>` at the top).

An alternative is to define local `kMPT_COMPACT_*` constants in `mpt_utility.h` (matching the `kMPT_` naming convention used for the other size constants) and keep the secp256k1 dependency internal. Minor either way — just flagging the convention inconsistency.

tesseract-ripple (commented):
Post-merge questions on a few API design choices. We had a parallel branch (PR #22) that made the same compact-sigma migration with some different decisions at the utility layer — wanted to understand the reasoning here so we can converge.

Nothing urgent, just want to make sure we're aligned before downstream (rippled) integration locks these in.

include/utility/mpt_utility.h
yinyiqian1: 1. This is to be consistent with other transactions. And it is more efficient to retrieve from ledger than deriving from priv.
2. amount_params's blinding factor is the same as the tx blinding factor to adapt to the new design, so we do not need to pass in the struct anymore. The amount and blinding factor can be found in the other passed in parameters.

include/utility/mpt_utility.h
yinyiqian1: Do you think we can extract the length constants into a shared file. We can add a `protocol` file shared by
both sdk layer `utility` and the other core files.

src/utility/mpt_utility.cpp
yinyiqian1: It looks like this was accidentally removed. I’ll add it back in a separate PR. In the meantime, rippled includes a workaround to implement the logic in the Attackthon branch. This function is still called in rippled feature branch and will be fixed after mpt-crypto new release.

include/utility/mpt_utility.h
yinyiqian1: Yes, in the previous version, I passed in the length because it varied whether there's an auditor or not. It is removed due to it's fixed length now.

But I’ll likely update the input to something like `proof[_some_length_]`, where _some_length_ equals `SECP256K1_COMPACT_STANDARD_PROOF_SIZE + kMPT_DOUBLE_BULLETPROOF_SIZE`, to stay consistent with other verification functions (e.g., `mpt_verify_clawback_proof`). But they are still pointers, but looks cleaner with an indicator of the length.

Currently, we enforce the length checks at the call sites. In rippled, we use a wrapper around the mpt_utility functions that prepares the inputs and validates parameters before passing them in.

Another option would be to include the length as a parameter in all functions. We can evaluate to see which approach we want to choose.

tesseract-ripple: this makes sense to me. i think keeping the length check at the call site is reasonable.

include/utility/mpt_utility.h
tesseract-ripple: i like that: #27 does this

PR #56, opened 2026-04-22
Add workflow to build shared libraries and attach to releases

Patel-Raj11:
Add `.github/workflows/build-shared-libs.yml`, a workflow that builds mpt-crypto as a shared library (`.dylib` / `.so` / `.dll`) for macOS (arm64, x86_64), Linux (arm64, x86_64, s390x), and Windows (x86_64), and attaches a single combined tarball to the corresponding GitHub release.

Downstream client libraries — xrpl4j (via JNA) and xrpl-py (via ctypes), and any future Rust/Go wrappers — need prebuilt mpt-crypto binaries they can load at runtime. Without this workflow, each consumer has to build mpt-crypto from source, which means every consumer repeats the Conan/CMake/OpenSSL/secp256k1 setup on its own CI. This workflow centralizes that work here.

### DIFF
--- a/.github/scripts/build-shared-lib.sh
+++ b/.github/scripts/build-shared-lib.sh
@@ -1,11 +1,4 @@
 #!/usr/bin/env bash
-# .github/scripts/build-shared-lib.sh
-#
-# Builds libmpt-crypto as a shared library with tests, via Conan + CMake.
-# Invoked by .github/workflows/build-shared-libs.yml — either directly on
-# a GitHub-hosted runner (Linux / macOS / Windows) or inside a
-# --platform linux/<arch> Docker container for platforms without native
-# runners (currently s390x via QEMU user-mode emulation).
 set -euo pipefail
 
 conan profile detect --force
--- a/.github/scripts/build-shared-lib.sh
+++ b/.github/scripts/build-shared-lib.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+# .github/scripts/build-shared-lib.sh
+#
+# Builds libmpt-crypto as a shared library with tests, via Conan + CMake.
+# Invoked by .github/workflows/build-shared-libs.yml — either directly on
+# a GitHub-hosted runner (Linux / macOS / Windows) or inside a
+# --platform linux/<arch> Docker container for platforms without native
+# runners (currently s390x via QEMU user-mode emulation).
+#
+# Preconditions: python3, conan, cmake, ninja (and MSVC on Windows) are
+# on PATH, and the cwd is the repo root. RUNNER_OS is read to decide
+# Windows-vs-POSIX behaviour; it's unset inside emulated containers, so
+# we default to Linux.
+set -euo pipefail
+
+conan profile detect --force
+conan remote add --index 0 --force xrplf https://conan.ripplex.io
+
+# fPIC is a POSIX-only concept; MSVC rejects it.
+CONAN_ARGS=(
+  -of build
+  --build=missing
+  -s build_type=Release
+  -o "&:shared=True"
+  -o "&:tests=True"
+  -o "secp256k1/*:shared=False"
+  -o "openssl/*:shared=False"
+)
+if [[ "${RUNNER_OS:-Linux}" != "Windows" ]]; then
+  CONAN_ARGS+=(-o "secp256k1/*:fPIC=True" -o "openssl/*:fPIC=True")
+fi
+conan install . "${CONAN_ARGS[@]}"
+
+# Windows uses the Visual Studio multi-config generator; everything else
+# uses Ninja (single-config, so CMAKE_BUILD_TYPE is baked in here).
+CMAKE_ARGS=(
+  -B build
+  -S .
+  -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake
+)
+if [[ "${RUNNER_OS:-Linux}" == "Windows" ]]; then
+  CMAKE_ARGS+=(
+    -G "Visual Studio 17 2022"
+    -A x64
+    -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON
+  )
+else
+  CMAKE_ARGS+=(
+    -G Ninja
+    -DCMAKE_BUILD_TYPE=Release
+  )
+fi
+cmake "${CMAKE_ARGS[@]}"
+
+cmake --build build --config Release
+
+# Windows has no rpath — test executables need the DLL's directory on
+# PATH. With the VS generator the DLL lands at build/Release/, so prepend
+# that. -C Release selects the VS multi-config build.
+pushd build > /dev/null
+CTEST_ARGS=(--output-on-failure)
+if [[ "${RUNNER_OS:-Linux}" == "Windows" ]]; then
+  export PATH="$(pwd)/Release:${PATH}"
+  CTEST_ARGS+=(-C Release)
+fi
+ctest "${CTEST_ARGS[@]}"
+popd > /dev/null
--- a/.github/scripts/build-shared-lib.sh
+++ b/.github/scripts/build-shared-lib.sh
@@ -6,17 +6,11 @@
 # a GitHub-hosted runner (Linux / macOS / Windows) or inside a
 # --platform linux/<arch> Docker container for platforms without native
 # runners (currently s390x via QEMU user-mode emulation).
-#
-# Preconditions: python3, conan, cmake, ninja (and MSVC on Windows) are
-# on PATH, and the cwd is the repo root. RUNNER_OS is read to decide
-# Windows-vs-POSIX behaviour; it's unset inside emulated containers, so
-# we default to Linux.
 set -euo pipefail
 
 conan profile detect --force
 conan remote add --index 0 --force xrplf https://conan.ripplex.io
 
-# fPIC is a POSIX-only concept; MSVC rejects it.
 CONAN_ARGS=(
   -of build
   --build=missing
@@ -31,8 +25,6 @@ if [[ "${RUNNER_OS:-Linux}" != "Windows" ]]; then
 fi
 conan install . "${CONAN_ARGS[@]}"
 
-# Windows uses the Visual Studio multi-config generator; everything else
-# uses Ninja (single-config, so CMAKE_BUILD_TYPE is baked in here).
 CMAKE_ARGS=(
   -B build
   -S .
@@ -54,9 +46,6 @@ cmake "${CMAKE_ARGS[@]}"
 
 cmake --build build --config Release
 
-# Windows has no rpath — test executables need the DLL's directory on
-# PATH. With the VS generator the DLL lands at build/Release/, so prepend
-# that. -C Release selects the VS multi-config build.
 pushd build > /dev/null
 CTEST_ARGS=(--output-on-failure)
 if [[ "${RUNNER_OS:-Linux}" == "Windows" ]]; then
--- a/.github/workflows/build-shared-libs.yml
+++ b/.github/workflows/build-shared-libs.yml
@@ -19,7 +19,13 @@
 #   darwin-x86-64/libmpt-crypto.dylib   (macOS x86_64)
 #   linux-aarch64/libmpt-crypto.so      (Linux ARM64)
 #   linux-x86-64/libmpt-crypto.so       (Linux x86_64)
+#   linux-s390x/libmpt-crypto.so        (Linux s390x, IBM Z — QEMU-built)
 #   win32-x86-64/mpt-crypto.dll         (Windows x86_64)
+#
+# Platforms without a native GitHub-hosted runner (s390x) are built inside a
+# --platform linux/<arch> Docker container via QEMU user-mode emulation.
+# The same .github/scripts/build-shared-lib.sh drives both the native and
+# the emulated builds — the workflow just decides where to run it.
 #
 # This workflow is invoked via `workflow_call` from build.yml — on every PR
 # and push to main (so the native-binary build path is exercised before a
@@ -65,89 +71,80 @@ jobs:
             platform: linux-x86-64
             lib_filename: libmpt-crypto.so
 
+          # ── Linux s390x (IBM Z) — built via QEMU user-mode emulation ──
+          # No native GHA runner exists for s390x. Builds and tests run
+          # inside a --platform linux/s390x container; QEMU emulates each
+          # instruction, so expect ~10-50x slower than native. Catches
+          # endianness bugs (s390x is big-endian) on every PR.
+          - os: ubuntu-latest
+            platform: linux-s390x
+            lib_filename: libmpt-crypto.so
+            docker_platform: linux/s390x
+            docker_image: s390x/ubuntu:24.04
+
           # ── Windows x86_64 ──
           - os: windows-latest
             platform: win32-x86-64
             lib_filename: mpt-crypto.dll
 
     runs-on: ${{ matrix.os }}
+    # Generous upper bound for the emulated s390x job (~30-60 min under
+    # QEMU when Conan has to build OpenSSL/secp256k1 from source).
+    # Native jobs finish in ~5-10 min and are unaffected.
+    timeout-minutes: 90
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      # ── QEMU for non-native Linux targets (s390x today) ──
+      - name: Set up QEMU
+        if: matrix.docker_platform
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
+        with:
+          platforms: ${{ matrix.docker_platform }}
+
+      # ── Toolchain setup for native runners. Skipped for Docker builds,
+      # which install the same tooling inside the container below. ──
       - name: Set up Python
+        if: ${{ !matrix.docker_platform }}
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: "3.12"
 
       - name: Install Conan and Ninja
+        if: ${{ !matrix.docker_platform }}
         run: pip install "conan>=2.0.0" ninja
 
       - name: Set up MSVC environment
         if: runner.os == 'Windows'
         uses: ilammy/msvc-dev-cmd@a102174a2b586eec2ea151a69e6fd14404a8ce7c # v1.13.0
 
-      - name: Configure Conan
+      # ── Build + test on the native runner (Linux/macOS/Windows) ──
+      - name: Build and test
+        if: ${{ !matrix.docker_platform }}
         shell: bash
-        run: |
-          conan profile detect --force
-          echo "==> Detected Conan profile:"
-          conan profile show
-          conan remote add --index 0 --force xrplf https://conan.ripplex.io
+        run: ./.github/scripts/build-shared-lib.sh
 
-      - name: Install dependencies
-        shell: bash
+      # ── Build + test inside a QEMU-emulated Linux container ──
+      # Everything — apt, pip, conan, cmake, ninja, the compiled binary,
+      # the test executables under ctest — runs translated through QEMU.
+      # Host/runner sees real big-endian ELF output under ./build/.
+      - name: Build and test in ${{ matrix.docker_platform }} container
+        if: matrix.docker_platform
         run: |
-          CONAN_ARGS=(
-            -of build
-            --build=missing
-            -s build_type=Release
-            -o "&:shared=True"
-            -o "&:tests=True"
-            -o "secp256k1/*:shared=False"
-            -o "openssl/*:shared=False"
-          )
-          if [[ "${RUNNER_OS}" != "Windows" ]]; then
-            CONAN_ARGS+=(-o "secp256k1/*:fPIC=True" -o "openssl/*:fPIC=True")
-          fi
-          conan install . "${CONAN_ARGS[@]}"
-
-      - name: Configure CMake
-        shell: bash
-        run: |
-          CMAKE_ARGS=(
-            -B build
-            -S .
-            -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake
-          )
-          if [[ "${RUNNER_OS}" == "Windows" ]]; then
-            CMAKE_ARGS+=(
-              -G "Visual Studio 17 2022"
-              -A x64
-              -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON
-            )
-          else
-            CMAKE_ARGS+=(
-              -G Ninja
-              -DCMAKE_BUILD_TYPE=Release
-            )
-          fi
-          cmake "${CMAKE_ARGS[@]}"
-
-      - name: Build shared library
-        shell: bash
-        run: cmake --build build --config Release
-
-      - name: Run tests
-        shell: bash
-        working-directory: build
-        run: |
-          CTEST_ARGS=(--output-on-failure)
-          if [[ "${RUNNER_OS}" == "Windows" ]]; then
-            export PATH="$(pwd)/Release:${PATH}"
-            CTEST_ARGS+=(-C Release)
-          fi
-          ctest "${CTEST_ARGS[@]}"
+          docker run --rm --platform ${{ matrix.docker_platform }} \
+            -v "${PWD}:/work" -w /work \
+            ${{ matrix.docker_image }} \
+            bash -c '
+              set -eux
+              apt-get update -qq
+              DEBIAN_FRONTEND=noninteractive apt-get install -qq -y \
+                python3 python3-pip \
+                cmake ninja-build \
+                build-essential pkg-config ca-certificates git
+              pip install --break-system-packages "conan>=2.0.0"
+              ./.github/scripts/build-shared-lib.sh
+            '
 
       - name: Stage shared library under output/<platform>/
         shell: bash
--- a/.github/workflows/build-shared-libs.yml
+++ b/.github/workflows/build-shared-libs.yml
@@ -30,7 +30,7 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 
+        description:
           "Release tag to build and attach binaries to (must already have a
           published GitHub release)"
         required: true
--- a/.github/workflows/build-shared-libs.yml
+++ b/.github/workflows/build-shared-libs.yml
@@ -12,7 +12,9 @@
 # attempt at full portability (no musl, no MT CRT).
 #
 # The output artifacts use CMake's natural library names under per-platform
-# subdirectories (OS × arch):
+# subdirectories (OS × arch), plus the public headers shared across platforms:
+#   include/secp256k1_mpt.h             (public API)
+#   include/utility/mpt_utility.h       (public API)
 #   darwin-aarch64/libmpt-crypto.dylib  (macOS ARM64)
 #   darwin-x86-64/libmpt-crypto.dylib   (macOS x86_64)
 #   linux-aarch64/libmpt-crypto.so      (Linux ARM64)
@@ -33,16 +35,6 @@ name: Build Shared Libraries
 
 on:
   workflow_call:
-    inputs:
-      ref:
-        description:
-          "Git ref (tag or SHA) to check out and build. Passed directly to
-          actions/checkout."
-        required: true
-        type: string
-
-permissions:
-  contents: read
 
 jobs:
   build:
@@ -80,8 +72,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          ref: ${{ inputs.ref }}
 
       # ── Set up build tools ──
       - name: Set up Python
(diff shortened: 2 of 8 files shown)


.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,275 @@
...
+      fail-fast: false
+      matrix:
+        include:
+          # ── macOS ARM64 (Apple Silicon) ──
+          - os: macos-14
+            platform: darwin-aarch64
+            lib_filename: libmpt-crypto.dylib
+
+          # ── macOS x86_64 (Intel) ──
+          - os: macos-15-intel
```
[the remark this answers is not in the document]
Patel-Raj11: @mathbunnyru  I am not sure if `macos15` runner provide x86_64 architecture? How can I check that?

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,275 @@
...
+            platform: darwin-x86-64
+            lib_filename: libmpt-crypto.dylib
+
+          # ── Linux ARM64 ──
+          - os: ubuntu-24.04-arm
```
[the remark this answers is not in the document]
Patel-Raj11: @mathbunnyru Are Github hosted runners like` ubuntu-24.04-arm` and `macos-15-intel` enabled in this repository? If not, then can we enable them, otherwise after switching to `heavy-arm64` runner, I won't be able to trigger the workflow on my fork as `heavy-arm64` won't be available there.

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,275 @@
...
+          TOOLCHAIN=$(find build -name "conan_toolchain.cmake" -print -quit)
+          if [ -z "${TOOLCHAIN}" ]; then
+            echo "ERROR: Could not find conan_toolchain.cmake under build/"
+            find build -type f -name "*.cmake" | head -20
+            exit 1
+          fi
+          echo "==> Using toolchain: ${TOOLCHAIN}"
+
+          cmake -B build -S . \
+            -G "Visual Studio 17 2022" \
```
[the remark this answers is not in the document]
Patel-Raj11: Addressed in [68cad80](https://github.com/XRPLF/mpt-crypto/pull/56/commits/68cad80d33e36ca8ca25b0e1941f4d42f8438bd7)

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,275 @@
...
+# ──────────────────────────────────────────────────────────────────────────────
+
+name: Build Shared Libraries
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to build and attach binaries to (must already have a
+          published GitHub release)"
```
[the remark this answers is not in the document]
Patel-Raj11: YAML double-quoted scalars apply line folding: a single newline inside the quotes becomes a space, and leading whitespace on continuation lines is stripped. So this:

description:
  "Release tag to build and attach binaries to (must already have a
  published GitHub release)"
is parsed as the single string "Release tag to build and attach binaries to (must already have a published GitHub release)" — no embedded newline, no extra spaces.

.github/workflows/build-shared-libs.yml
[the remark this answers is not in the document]
mathbunnyru: I think copilot is wrong here, `macos-15-intel` definitely exists: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

.github/workflows/build-shared-libs.yml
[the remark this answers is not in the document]
mathbunnyru: They are enabled, it's GitHub provided runners

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,277 @@
...
+#   win32-x86-64/mpt-crypto.dll         (Windows x86_64)
+#
+# A single combined archive (mpt-crypto-natives-<tag>.tar.gz) is attached to
+# the release.
+# ──────────────────────────────────────────────────────────────────────────────
+
+name: Build Shared Libraries
+
+on:
+  workflow_dispatch:
```
mathbunnyru: This is a bad design - we'll only know that this workflow is correct on release, which is way too late.
And if we didn't forget to run it
I'll explain in the general message how to do it better

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,277 @@
...
+    inputs:
+      tag:
+        description:
+          "Release tag to build and attach binaries to (must already have a
+          published GitHub release)"
+        required: true
+        type: string
+
+permissions:
+  contents: read
```
mathbunnyru: Let's add permissions to every job separate, because this looks like all jobs have only `contents: read` permission

mathbunnyru (changes requested):
I propose another architecture, 2 jobs here, make this workflow callable and have 2 jobs:
1. build
2. bundle (depends on build)

In `build.yml` workflow you just call this reusable workflow.

Also, in `build.yml` in the end add creating a release, if it's tag being pushed, which would upload the bundle. 

What this achieves:
1. We always test workflows are actually in a working state in PRs
2. We publish release automatically, on tag push, instead of creating manually
3. It would be impossible to forget to add artifacts later
4. Instead of adding them at a later stage, we publish them with the release creation
5. This allows to enable immutable releases (a good security feature of GitHub)

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,277 @@
...
+
+  # ──────────────────────────────────────────────────────────────────────────
+  # Bundle all platform artifacts into a single downloadable archive.
+  #   darwin-aarch64/libmpt-crypto.dylib
+  #   darwin-x86-64/libmpt-crypto.dylib
+  #   linux-aarch64/libmpt-crypto.so
+  #   linux-x86-64/libmpt-crypto.so
+  #   win32-x86-64/mpt-crypto.dll
+  # ──────────────────────────────────────────────────────────────────────────
+  bundle:
```
pdp2121, line 148: Can we also include these in the tarball:
```
include/secp256k1_mpt.h
include/utility/mpt_utility.h
```
As they are canonical API blueprint, and would be required for some bindings (xrpl-py cffi would need them)?

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      ref:
+        description:
+          "Git ref (tag or SHA) to check out and build. Passed directly to
+          actions/checkout."
+        required: true
+        type: string
```
mathbunnyru: I don't think you need these, and you can just use actions/checkout with default params

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+        shell: bash
+        run: |
+          conan install . \
+            -of build \
+            --build=missing \
+            -s build_type=Release \
+            -o "&:shared=True" \
+            -o "&:tests=False" \
+            -o "secp256k1/*:shared=False" \
+            -o "openssl/*:shared=False"
```
mathbunnyru: These are almost fully the same. Please merge them, if there are differences, use the env variable(s) passed to a step

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+
+      # ── Configure CMake ──
+      - name: Configure CMake (Unix)
```
mathbunnyru: Also please merge for Windows / non-Windows

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+        if: runner.os != 'Windows'
+        shell: bash
+        run: |
+          TOOLCHAIN=$(find build -name "conan_toolchain.cmake" -print -quit)
```
mathbunnyru: Location is deterministic, please take a look at how build.yml does it

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+          if [ -z "${TOOLCHAIN}" ]; then
+            echo "ERROR: Could not find conan_toolchain.cmake under build/"
+            find build -type f -name "*.cmake" | head -20
+            exit 1
+          fi
+          echo "==> Using toolchain: ${TOOLCHAIN}"
```
mathbunnyru: Not needed

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+      # ── Stage ──
+      # Locate the built library (path varies by generator: build/ for Ninja,
+      # build/Release/ for Visual Studio) and place it under output/<platform>/
+      # so upload-artifact preserves the platform subdirectory.
+      - name: Stage shared library under output/<platform>/
+        shell: bash
+        run: |
+          mkdir -p output/${{ matrix.platform }}
+
+          BUILT_LIB=$(find build -maxdepth 2 -name "${{ matrix.lib_filename }}" -print -quit)
```
mathbunnyru: Don't use find, location is deterministic, use env for step if there are differences for Windows / non-Windows

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+
+          if [ -z "${BUILT_LIB}" ]; then
+            echo "ERROR: Could not find ${{ matrix.lib_filename }} in build/"
+            echo "Build directory contents:"
+            find build -type f \( -name "*.dylib" -o -name "*.so" -o -name "*.dll" \) 2>/dev/null
+            exit 1
+          fi
```
mathbunnyru: Not needed, cp later will fail

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+
+          echo "==> Found: ${BUILT_LIB}"
+          cp "${BUILT_LIB}" "output/${{ matrix.platform }}/"
+          echo "==> Packaged: output/${{ matrix.platform }}/${{ matrix.lib_filename }}"
+
+      # ── Upload per-platform artifact ──
```
mathbunnyru: Please, remove such comments if they don't provide additional information

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+          path: output/
+          if-no-files-found: error
+
+  # ──────────────────────────────────────────────────────────────────────────
+  # Bundle all platform artifacts into a single downloadable archive.
+  #   darwin-aarch64/libmpt-crypto.dylib
+  #   darwin-x86-64/libmpt-crypto.dylib
+  #   linux-aarch64/libmpt-crypto.so
+  #   linux-x86-64/libmpt-crypto.so
+  #   win32-x86-64/mpt-crypto.dll
```
mathbunnyru: This duplicated the list above. I would remove this part

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+          path: natives/
+          merge-multiple: true
+
+      - name: Display bundled structure
+        run: |
+          echo "==> Bundled native libraries:"
+          find natives/ -type f | sort
+          echo ""
+          echo "==> File sizes:"
+          find natives/ -type f -exec ls -lh {} \;
```
mathbunnyru: This all can probably be one `ls` command

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,247 @@
...
+
+      - name: Create combined archive
+        run: |
+          tar -czf mpt-crypto-natives.tar.gz -C natives .
+          echo "==> Created mpt-crypto-natives.tar.gz"
```
mathbunnyru: I think these commands are redundant in most places, since steps already describe what was done

.github/workflows/build.yml
```diff
@@ -90,3 +95,52 @@ jobs:
...
+    with:
+      ref: ${{ github.sha }}
+
+  # On tag push only: publish a GitHub release and attach the native bundle.
+  release:
+    name: Publish release
+    # Run on tag push when native-binaries succeeded, and either build-and-test
+    # passed (upstream) or was skipped (fork). `!failure() && !cancelled()`
+    # is the GH Actions idiom to override the default "skip if any need was
+    # skipped" behavior without also running on real failures.
```
mathbunnyru: I don't think we need to care about release on forks, since the binaries still won't be there. So, unless there is a reason you have to do it this way, I would keep it simple (without `if`)

.github/workflows/build.yml
```diff
@@ -90,3 +95,52 @@ jobs:
...
+
+      - name: Create GitHub release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [ ! -f mpt-crypto-natives.tar.gz ]; then
+            echo "ERROR: mpt-crypto-natives.tar.gz not found in downloaded bundle"
+            ls -la
+            exit 1
+          fi
```
mathbunnyru: No need for this, `mv` will fail later

mathbunnyru (changes requested):

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: I have removed the echo statements altogether as it provides little value.

.github/workflows/build-shared-libs.yml
Patel-Raj11: Removed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Removed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a) by extracting out common flags. Here it is more cleaner than using env variables to distinguish between flags for Windows/Non Windows

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Added in [b84f72f](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b84f72f1107ff3c026a34ed96b958c356ee7689a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: I removed the permissions block altogether as it is no more needed. The default repository provided permissions suffice.

.github/workflows/build-shared-libs.yml
Patel-Raj11: Thanks for the suggestions. I have incorporated the changes as suggested. Apart from these I have made changes in the following tests files as windows has never compiled this project's tests before. Turning on `tests=True` in the new shared-libs Windows build surfaced them.

1. tests/CMakeLists.txt
What changed: `secp256k1` → `secp256k1::secp256k1` in every DEPENDENCIES list.
On MSVC, secp256k1 is translated to the literal filename secp256k1.lib and searched directly — but the file Conan actually produces is libsecp256k1.lib (with the lib prefix). Result: LNK1181: cannot open input file 'secp256k1.lib'. 

2. removed `CLOCK_MONOTONIC` in `test_bulletproof_agg.c`
`CLOCK_MONOTONIC` is a POSIX constant, defined in <time.h> only on POSIX systems (Linux, macOS, BSD). It is not part of standard C, and MSVC's <time.h> does not define it. Replaced every clock_gettime(CLOCK_MONOTONIC, &ts) call with timespec_get(&ts, TIME_UTC).

3. tests/test_same_plaintext_multi_shared_r.c
The array unsigned char proof[proof_len] is a C99 Variable-Length Array. MSVC doesn't support it. Used malloc/free instead.

.github/workflows/build.yml
Patel-Raj11: Removed it here - [3547166](https://github.com/XRPLF/mpt-crypto/pull/56/commits/35471663b1ece292724c7311f2845b7432780166)
I added that to make sure that release [note](https://github.com/Patel-Raj11/mpt-crypto/releases/tag/0.2.0) gets generated in my fork with artifacts attached.

.github/workflows/build.yml
Patel-Raj11: Removed here [3547166](https://github.com/XRPLF/mpt-crypto/pull/56/commits/35471663b1ece292724c7311f2845b7432780166)

.github/workflows/build-shared-libs.yml
mathbunnyru: A good practice is to be explicit and tell which permissions are needed

.github/workflows/build-shared-libs.yml
mathbunnyru: Nice, thanks!

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,200 @@
...
+          - os: windows-latest
+            platform: win32-x86-64
+            lib_filename: mpt-crypto.dll
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      # ── Set up build tools ──
```
mathbunnyru: I would suggest removing this comment, test and stage comments as well

mathbunnyru (commented):
Looks so much better to me, thanks!
Only 2 things left - let's add correct permissions, and remove redundant comments

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,200 @@
...
+
+on:
+  workflow_call:
+
+jobs:
+  build:
+    name: Build ${{ matrix.platform }}
+    strategy:
+      fail-fast: false
+      matrix:
```
placintaalexandru, line 52: Please also add support for `s390x`

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed [9683245](https://github.com/XRPLF/mpt-crypto/pull/56/commits/9683245e114d5daf08f30bac7e81e1243affbb5b)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed [9683245](https://github.com/XRPLF/mpt-crypto/pull/56/commits/9683245e114d5daf08f30bac7e81e1243affbb5b)

.github/workflows/build.yml
```diff
@@ -90,3 +95,34 @@ jobs:
...
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: mpt-crypto-natives-bundle
+
+      - name: Create GitHub release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          ARCHIVE="mpt-crypto-natives-${GITHUB_REF_NAME}.tar.gz"
+          mv mpt-crypto-natives.tar.gz "${ARCHIVE}"
```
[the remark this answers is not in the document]
Patel-Raj11: `mv mpt-crypto-natives.tar.gz "${ARCHIVE}"` doesn't fail and artifacts gets properly attached to release notes - https://github.com/Patel-Raj11/mpt-crypto/releases/tag/0.2.0

mathbunnyru (approved):
LGTM, but please take a look at copilot comment

.github/workflows/build-shared-libs.yml
Patel-Raj11: Added the support for s390x here [5d63166](https://github.com/XRPLF/mpt-crypto/pull/56/commits/5d63166608551ffe10a4b18657756309de065257)

Patel-Raj11: @mathbunnyru I have added support for `s390x` architecture, and it uses QEMU emulation using docker container. So there are some refactors involved in the build matrix. Can you have a look once again? Thanks!

tests/test_bulletproof_agg.c
```diff
@@ -57,26 +57,26 @@ void run_test_case(secp256k1_context *ctx, const char *name, uint64_t *values,
...
   struct timespec t_p_start, t_p_end;
-  clock_gettime(CLOCK_MONOTONIC, &t_p_start);
+  timespec_get(&t_p_start, TIME_UTC);
 
   EXPECT(secp256k1_bulletproof_prove_agg(ctx, proof, &proof_len, values,
                                          (const unsigned char *)blindings,
                                          num_values, &pk_base, context_id));
 
-  clock_gettime(CLOCK_MONOTONIC, &t_p_end);
+  timespec_get(&t_p_end, TIME_UTC);
```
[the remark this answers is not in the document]
Patel-Raj11: These timings are purely printf output. No EXPECT() check and no CI gate depends on the elapsed value. So the complexity of adding the shim for tests that run on windows (`QueryPerformanceCounter/GetSystemTimePreciseAsFileTime`) wasn't buying anything the tests actually need.

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,182 @@
...
+            platform: win32-x86-64
+            lib_filename: mpt-crypto.dll
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set up QEMU
+        if: matrix.docker_platform
```
mathbunnyru: Maybe?
```suggestion
        if: matrix.docker_platform != ""
```

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,182 @@
...
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
```
mathbunnyru: There is v4.0.0, please update.

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,182 @@
...
+          platforms: ${{ matrix.docker_platform }}
+
+      - name: Set up Python
+        if: ${{ !matrix.docker_platform }}
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.12"
+
+      - name: Install Conan and Ninja
+        if: ${{ !matrix.docker_platform }}
```
mathbunnyru:
```suggestion
        if: matrix.docker_platform == ""
```

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,182 @@
...
+        run: pip install "conan>=2.0.0" ninja
```
mathbunnyru: I don't think we need to specify `conan>=2.0.0`, doesn't it install quite recent version by default?

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,182 @@
...
+
+      - name: Set up MSVC environment
+        if: runner.os == 'Windows'
+        uses: ilammy/msvc-dev-cmd@a102174a2b586eec2ea151a69e6fd14404a8ce7c # v1.13.0
```
mathbunnyru, line 109: Btw, this action is out of date, maybe we can use XRPLF's windows runners and we won't need this actions at all?
It's `windows` runner tag if I remember correctly

.github/scripts/build-shared-lib.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+# .github/scripts/build-shared-lib.sh
+#
+# Builds libmpt-crypto as a shared library with tests, via Conan + CMake.
+# Invoked by .github/workflows/build-shared-libs.yml — either directly on
+# a GitHub-hosted runner (Linux / macOS / Windows) or inside a
+# --platform linux/<arch> Docker container for platforms without native
+# runners (currently s390x via QEMU user-mode emulation).
```
mathbunnyru: I don't think this comment is needed, this script just always natively builds the shared lib, no matter where it's been called from

.github/scripts/build-shared-lib.sh
```diff
@@ -0,0 +1,56 @@
...
+CONAN_ARGS=(
+  -of build
+  --build=missing
+  -s build_type=Release
+  -o "&:shared=True"
+  -o "&:tests=True"
+  -o "secp256k1/*:shared=False"
+  -o "openssl/*:shared=False"
+)
+if [[ "${RUNNER_OS:-Linux}" != "Windows" ]]; then
```
mathbunnyru, line 16: Why do we need the `Linux` default when using this variable everywhere?

.github/workflows/build-shared-libs.yml
```diff
@@ -0,0 +1,182 @@
...
+                cmake ninja-build \
+                build-essential pkg-config ca-certificates git
+              pip install --break-system-packages "conan>=2.0.0"
+              ./.github/scripts/build-shared-lib.sh
+            '
+
+      - name: Stage shared library under output/<platform>/
+        shell: bash
+        env:
+          LIB_DIR: ${{ runner.os == 'Windows' && 'build/Release' || 'build' }}
```
mathbunnyru, line 136: I see you used my advice to sometimes use `env` variables :)

mathbunnyru (changes requested):

.github/workflows/build-shared-libs.yml
Patel-Raj11: It would stay undefined for other platforms and this step would be skipped if we keep using `if: matrix.docker_platform`. I think we can keep this as it is.

.github/workflows/build-shared-libs.yml
Patel-Raj11: If I use XRPLF's `windows` runner, I won't be able to test my change in the fork. Could you add me as a member to this repository so that my commits auto-triggers the workflows?

.github/scripts/build-shared-lib.sh
Patel-Raj11: When the script runs inside the docker container, `RUNNER_OS` is not set. So we are defaulting it to Linux here since it is always `s390x/ubuntu:24.04` image.

.github/scripts/build-shared-lib.sh
Patel-Raj11: Addressed in [b86ce13](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b86ce133506e6eebbcc8624e6830fa3e47ca1c4a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Yes it defaults to recent version that is greater than 2.0.0. Addressed in [b86ce13](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b86ce133506e6eebbcc8624e6830fa3e47ca1c4a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b86ce13](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b86ce133506e6eebbcc8624e6830fa3e47ca1c4a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed in [b86ce13](https://github.com/XRPLF/mpt-crypto/pull/56/commits/b86ce133506e6eebbcc8624e6830fa3e47ca1c4a)

.github/workflows/build-shared-libs.yml
Patel-Raj11: Addressed this as well, to be consistent with `if: matrix.docker_platform == ""` cases.

.github/workflows/build-shared-libs.yml
mathbunnyru: When we merge your first contribution, it should work fine

.github/scripts/build-shared-lib.sh
mathbunnyru: Thanks, makes sense

mathbunnyru (approved):
LGTM, but I will ask to make further improvements

PR #57, opened 2026-04-27
Use XRPLF heavy runners

Patel-Raj11:
## Changes
- Use heavy / heavy-arm64 for Linux builds (inside the `clio-ci` container, matching `build-and-test`)
- Use macos15 for macOS arm64 builds
- Use windows runner for Windows builds; remove explicit MSVC environment setup (`ilammy/msvc-dev-cmd`)
- Replace setup-python + pip install with `XRPLF/actions/prepare-runner`
- Add `Cleanup workspace` step for the persistent macos15 runner
- Skip `native-binaries` on forks (matches `build-and-test`'s pattern)

## Still on GHA-hosted runners
- **macOS Intel** (`macos-15-intel`) — no XRPLF Intel runner available
- **Linux s390x** (`ubuntu-latest`) — QEMU setup needs privileged binfmt registration, which heavy doesn't allow.

### DIFF
--- a/.github/workflows/build-shared-libs.yml
+++ b/.github/workflows/build-shared-libs.yml
@@ -94,15 +94,11 @@ jobs:
         with:
           platforms: ${{ matrix.docker_platform }}
 
-      - name: Set up Python
+      - name: Prepare runner
         if: matrix.docker_platform == ''
-        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        uses: XRPLF/actions/prepare-runner@2cbf481018d930656e9276fcc20dc0e3a0be5b6d
         with:
-          python-version: "3.12"
-
-      - name: Install Conan and Ninja
-        if: matrix.docker_platform == ''
-        run: pip install conan ninja
+          enable_ccache: false
 
       - name: Build and test
         if: matrix.docker_platform == ''
--- a/.github/workflows/build-shared-libs.yml
+++ b/.github/workflows/build-shared-libs.yml
@@ -52,7 +52,7 @@ jobs:
       matrix:
         include:
           # ── macOS ARM64 (Apple Silicon) ──
-          - os: macos-14
+          - os: macos15
             platform: darwin-aarch64
             lib_filename: libmpt-crypto.dylib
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-shared-libs.yml
```diff
@@ -62,14 +62,16 @@ jobs:
...
 
           # ── Linux x86_64 ──
-          - os: ubuntu-latest
+          - os: heavy
             platform: linux-x86-64
             lib_filename: libmpt-crypto.so
+            container: '{ "image": "ghcr.io/xrplf/clio-ci:14342e087ceb8b593027198bf9ef06a43833c696" }'
 
           # ── Linux s390x (IBM Z) — built via QEMU user-mode emulation ──
           - os: ubuntu-latest
```
mathbunnyru, line 77: We can and should still use heavy here

mathbunnyru (changes requested):

.github/workflows/build-shared-libs.yml
```diff
@@ -62,14 +62,16 @@ jobs:
             lib_filename: libmpt-crypto.dylib
 
           # ── Linux ARM64 ──
-          - os: ubuntu-24.04-arm
+          - os: heavy-arm64
             platform: linux-aarch64
             lib_filename: libmpt-crypto.so
+            container: '{ "image": "ghcr.io/xrplf/clio-ci:14342e087ceb8b593027198bf9ef06a43833c696" }'
```
mathbunnyru, line 68: I'm not sure how I feel about using Clio images

Patel-Raj11: `build-and-test` uses Clio images for Linux runners. So I have followed the same convention here. I think it has C/C++ toolchains, CMake, Conan, Python etc already installed. Do you suggest we don't use it and have the toolchain setup inside the workflow itself as an extra step?

mathbunnyru: Ah, I was lazy back then, let's keep as is for now :(

.github/workflows/build-shared-libs.yml
Patel-Raj11: @mathbunnyru I tried using heavy runners for s390x, but it keeps failing with the error seen in below mentioned run. After some back and forth with Claude, it seems that heavy runners are running in some `sandbox/VM/namespace` such that `docker/setup-qemu-action` is not working as intended. 
  
https://github.com/XRPLF/mpt-crypto/actions/runs/24976270406/job/73128730834

Can you see if you can suggest something based on the error message? Or s390x is quite unique (it needs docker emulation) so can we keep using GHA runners?

Issue #39, opened 2026-04-13
[Bug] compute_amount_point returns 0 for amount=0, breaking clawback proofs


mvadari:
## Claude Sonnet 4.6 — BUG-02

**Location:** `src/mpt_internal.h:78–89` (`compute_amount_point`), called in:
- `src/proof_compact_clawback.c:122–124`
- `src/proof_link.c` (implicitly through mG handling)

**Description:**
`compute_amount_point` explicitly returns 0 (failure) when `amount == 0`:
```c
if (amount == 0)
    return 0;
```
In `secp256k1_compact_clawback_prove`:
```c
if (!compute_amount_point(ctx, &mG, amount))
    goto cleanup;
```
`cleanup` is reached with `ok = 0`, so the function returns 0 unconditionally when amount=0. The clawback verifier has the same failure:
```c
if (!compute_amount_point(ctx, &mG, amount))
    return 0;
```
A zero-amount clawback is semantically valid (issuer reclaims 0 tokens), and the protocol equation `C2 - 0*G = sk_iss * C1` (which reduces to `C2 = sk_iss * C1`) is well-defined. Both prover and verifier refuse to function.

**Impact:** Clawback proofs for zero-amount transactions are entirely impossible. **Critical.**

---

## Augment Code — Bug 8

**Location:** `src/mpt_internal.h:83-84`, called from `src/proof_compact_clawback.c:122, 252`
**Description:** `compute_amount_point` returns 0 when `amount == 0`. Both the clawback prover (line 122) and verifier (line 252) call this function and treat a 0 return as failure. While clawing back 0 is arguably pointless, the SPEC's validation rules say "The revealed amount m is non-negative and well-formed" without explicitly excluding 0. If a validator receives a clawback with m=0, the crypto code silently rejects what might be a valid (if pointless) transaction.
**Impact Level:** Low — clawback of 0 is semantically vacuous; likely blocked at higher validation layer.

tesseract-ripple:
Fixed by #59 (commit a70e782a, merged 2026-04-27). The PR added explicit `if (amount == 0)` branches at every call site of `compute_amount_point` that needs to handle clawback / encryption with zero amount. Verified against current `main`:

- `src/elgamal.c:96-108` (encrypt) — branches on `amount == 0` and sets `*c2 = S`
- `src/elgamal.c:334-346` (verify_encryption) — branches and sets `expected_c2 = S`
[...]

Issue #38, opened 2026-04-13
[Bug] compute_sigma_response fails when secret witness is zero (amount=0 or balance=0)


mvadari:
## Claude Sonnet 4.6 — BUG-01

**Location:** `src/mpt_internal.h:94–116` (`compute_sigma_response`), propagates to:
- `src/proof_compact_standard.c:347,350,356` (z_m when amount=0, z_b when balance=0)
- `src/proof_link.c:206–208` (sm when amount=0)
- `src/proof_same_plaintext_multi_shared_r.c:227–229` (s_m when amount=0)
- `src/proof_compact_convertback.c:244` (z_b when balance=0)

**Description:**
`compute_sigma_response` computes `z = nonce + e * secret` by calling:
```c
memcpy(term, secret, 32);
if (!secp256k1_ec_seckey_tweak_mul(ctx, term, e))
    return 0;
```
When `secret = 0` (i.e., `m_scalar` or `b_scalar` is all-zeros for a zero amount/balance), `secp256k1_ec_seckey_tweak_mul` is called with `term = 0`. The secp256k1 API rejects 0 as a secret key argument, so the function returns 0, propagating failure to the caller. The mathematical result (`0 * e = 0`, giving `z = nonce + 0 = nonce`) is perfectly valid — the implementation wrongly treats the zero witness as an invalid secret key.

**Impact:** Proof generation fails entirely for any transaction involving a zero amount (send) or zero balance (convertback, compact standard). Zero-value balances and transfers are legitimate protocol states. **Critical.**

---

## Augment Code — Bug 1

**Location:** `src/mpt_internal.h:104`, called from `proof_compact_standard.c:356`, `proof_compact_convertback.c:244`, `proof_link.c:206`
**Description:** The `compute_sigma_response` helper uses `secp256k1_ec_seckey_tweak_mul(ctx, term, e)` where `term` is a copy of the secret witness. `secp256k1_ec_seckey_tweak_mul` rejects zero inputs (returns 0 if the secret key is zero). When `balance == 0` (sender spends entire balance) or `amount == 0`, the corresponding scalar is all-zeros and the response computation fails. Mathematically, the response `z = nonce + e*0 = nonce` is perfectly valid, but the implementation cannot compute it. This means:
- **Compact Standard Send:** Cannot prove `balance == 0` (sender spent their entire balance). The `z_b = epsilon + e*0` computation fails.
[...]

PR #58, opened 2026-04-27
fix: make compute_sigma_response infallible on zero witness

tesseract-ripple:
## Summary

Fixes the zero-witness issue in `compute_sigma_response` (tracks #38) and tightens negative-path assertions in the aggregated bulletproof tests.

## The bug (issue #38)

`compute_sigma_response` previously used `secp256k1_ec_seckey_tweak_mul` / `secp256k1_ec_seckey_tweak_add` to compute `z = nonce + e * secret (mod n)`. Those APIs call `seckey_verify` internally and reject scalars that are zero or ≥ group order — so whenever the witness (e.g. amount `m = 0` or balance remainder `v = 0`) is zero, the intermediate `e * secret` term is zero and the call fails. The prover then aborts on a legal zero-valued statement.

## Fix

Switch to the internal scalar-field helpers `secp256k1_mpt_scalar_mul` / `secp256k1_mpt_scalar_add`, which operate on full 32-byte field elements and never reject zero. `compute_sigma_response` becomes infallible (`void`), so all 10 call sites across the four sigma-proof variants drop their error-handling branches:

- `proof_compact_standard.c` (5 sites: `z_r`, `z_m`, `z_sk`, `z_rb`, `z_b`)
- `proof_compact_convertback.c` (3 sites: `z_b`, `z_sk`, `z_rho`)
- `proof_compact_clawback.c` (1 site: `z_sk`)
- `proof_pok_sk.c` (1 site: `s`)

### New tests

- `tests/test_compact_standard.c` — factored into a parameterized case-runner; adds *zero amount* (`m = 0`) and *zero remainder* (`v = amount = balance`) cases on top of the primary non-zero case.
- `tests/test_compact_convertback.c` — factored similarly; adds *zero balance* case.
- Both rely on a small `value_times_g` / `uint64_to_scalar32` helper added to `tests/test_utils.h`, which skips the `v*G` term when `v = 0` (libsecp's public API cannot emit the point at infinity).

Full `ctest`: 13/13 pass with `-Werror`.

## Second commit: ToB April 2026 fix-review p-66 caveat (RIPCTX-6)

The ToB fix-review flagged that the aggregated-bulletproof negative tests were only checking *absence of crash*, not *rejection*. This commit:

1. Replaces the prior `if (ok) { fprintf; exit; }` pattern with `EXPECT(ok == 0)` for the tampered-value case.
2. Adds a second negative test (*same value, different blinding*) that exercises the rejection path even when the value witness itself is 0 (no G-term in the original commitment).

## Follow-up (not in this PR)

Zero-amount handling in the Fiat-Shamir transcripts + clawback path (#39) is a separate concern — tracked for a follow-up PR.

## Test plan

- [x] `ctest` passes 13/13 with `-Werror`
- [x] Zero-amount, zero-remainder, zero-balance cases exercise the new code path specifically
- [x] Pre-commit hooks pass (clang-format, gersemi, etc.)

### DIFF
--- a/src/mpt_internal.h
+++ b/src/mpt_internal.h
@@ -29,9 +29,13 @@
             return 0;       \
     } while (0)
 
-/* Forward-declare secp256k1_mpt helpers used in nonce generation. */
+/* Forward-declare secp256k1_mpt scalar helpers. */
 void
 secp256k1_mpt_scalar_reduce32(unsigned char out32[32], unsigned char const in32[32]);
+void
+secp256k1_mpt_scalar_add(unsigned char* res, unsigned char const* a, unsigned char const* b);
+void
+secp256k1_mpt_scalar_mul(unsigned char* res, unsigned char const* a, unsigned char const* b);
 
 /** Returns 1 if pk1 == pk2, 0 otherwise. */
 static inline int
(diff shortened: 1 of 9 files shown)


mrtcnk (approved):
LGTM!

PR #59, opened 2026-04-27
fix: handle zero amount in clawback Fiat-Shamir transcript

tesseract-ripple:
## Summary

Fixes the zero-amount abort in the compact clawback proof. Tracks issue #39. Also responds to the broader zero-handling theme in the ToB code-quality findings (Appendix B, "Code Quality Issues" — discrepancy between implementation and specification for zero-handling).

This PR is independent of the companion PR #58 (sigma-response zero witness) — they touch disjoint code and can merge in either order.

## The bug (issue #39 / internal S9/I9)

In `proof_compact_clawback.c`, both the prover and the verifier unconditionally call `compute_amount_point(ctx, &mG, amount)` to obtain the statement point `m*G`. That helper returns 0 when `amount == 0` (libsecp256k1 cannot represent the point at infinity), so the prove/verify flow aborts on legitimate zero-balance clawback ciphertexts.

The Fiat-Shamir transcript additionally serializes `m*G` as 33 compressed bytes. When `amount = 0`, there is no point to serialize — the internal S9/I9 note recommends substituting 33 zero bytes in this position so the transcript remains well-defined.

## Fix

**Unified transcript helper** — new static function `digest_update_amount_point()` encapsulates the convention in exactly one place:

- `amount > 0`: feed the 33-byte compressed encoding of `m*G`.
- `amount == 0`: feed 33 zero bytes.

33 zeros is not a valid compressed-point encoding (valid prefixes are `0x02` / `0x03`), so the sentinel is unambiguous and cannot collide with any real `m*G`.

`compute_compact_clawback_challenge()` now takes `uint64_t amount` instead of `const secp256k1_pubkey *mG`. Both prover and verifier pass `amount` directly, so the transcript convention can never drift between the two sides.

**Prover**
- Drops the unconditional `compute_amount_point()` call (and the `mG` local) at the top of the function.
- The inline `stmt_hash` used for deterministic-nonce generation now calls the same `digest_update_amount_point()` helper, so prover-side nonce derivation is also well-defined at `amount = 0`.

**Verifier**
- Drops the unconditional `compute_amount_point()` call.
- Reconstructs `T2 = z_sk*C1 - e*(C2 - m*G)` by short-circuiting `C2 - m*G` to `C2` when `amount == 0` (the `-m*G` term is the point at infinity).

## Why not change `equality_proof.c` too?

`equality_proof.c` currently handles `amount = 0` by **omitting** the `m*G` term from the hash (variable-length transcript). Internal S9/I9 recommends the same 33-zero-bytes convention there. That file is slated for removal in #25, so this PR does not change it — touching a soon-to-be-deleted transcript would create pointless merge noise. If #25 stalls, a follow-up can align `equality_proof.c` with the convention chosen here.

## For team discussion

I picked **33 zero bytes** as the sentinel for "no `m*G`" in the transcript. Alternatives considered:

| Option | Transcript bytes for m=0 | Notes |
|---|---|---|
| **33 zero bytes (chosen)** | fixed | Matches internal S9/I9 recommendation. Fixed-length, unambiguous (not a valid compressed encoding). |
| Omit the term | variable | What `equality_proof.c` does today. Variable-length hash inputs are a mild footgun. |
| 1-byte domain-separator sentinel | fixed (1) | Non-standard; no precedent in this codebase. |

Happy to revisit if anyone prefers a different convention — this PR is the first time it's been applied in mpt-crypto and sets precedent.

## Test plan

- [x] \`tests/test_compact_clawback.c\` refactored into a parameterized case-runner.
- [x] **Primary case** (amount = 500000): full negative-path battery passes.
- [x] **Zero-amount case** (amount = 0): prove + verify succeed; full negatives pass. The "wrong amount" negative becomes \`amount + 1 = 1\`, which specifically exercises the transcript-distinguishability boundary between the 33-zero sentinel and a serialized \`1*G\`.
- [x] Full \`ctest\`: 13/13 pass.
- [x] Pre-commit hooks pass (clang-format, etc.).

### DIFF
--- a/src/proof_compact_clawback.c
+++ b/src/proof_compact_clawback.c
@@ -31,12 +31,39 @@
 
 static const char DOMAIN_COMPACT_CLAWBACK[] = "CMPT_CLAWBACK_SIGMA";
 
+/* Feed m*G into an SHA-256 stream as 33 compressed bytes. When amount == 0,
+ * libsecp256k1 cannot represent (and therefore cannot serialize) the point
+ * at infinity, so we substitute 33 zero bytes. 33 zeros is not a valid
+ * compressed-point encoding (the valid prefixes are 0x02 and 0x03), so the
+ * sentinel is unambiguous and cannot collide with a real m*G. Prover and
+ * verifier must both use this helper so the transcript stays consistent. */
+static int digest_update_amount_point(const secp256k1_context *ctx,
+                                      EVP_MD_CTX *mdctx, uint64_t amount)
+{
+  unsigned char buf[33];
+  if (amount == 0)
+  {
+    memset(buf, 0, 33);
+  }
+  else
+  {
+    secp256k1_pubkey mG;
+    size_t len = 33;
+    if (!compute_amount_point(ctx, &mG, amount))
+      return 0;
+    if (!secp256k1_ec_pubkey_serialize(ctx, buf, &len, &mG,
+                                       SECP256K1_EC_COMPRESSED) ||
+        len != 33)
+      return 0;
+  }
+  return EVP_DigestUpdate(mdctx, buf, 33);
+}
+
 static int compute_compact_clawback_challenge(
     const secp256k1_context *ctx, unsigned char *e_out,
     const secp256k1_pubkey *P_iss, const secp256k1_pubkey *C1,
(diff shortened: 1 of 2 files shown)


mrtcnk (approved):
LGTM!

mrtcnk (approved):
LGTM!

Issue #115, opened 2026-06-02
[API Hardening] Avoid partial state changes before failure in utility wrappers and elgamal


tesseract-ripple:
**Severity:** Informational; defensive.
**Source:** TOB May 2026 revision, Appendix E — "Partial State
Changes Before Failure" (pp. 66-68). mpt-crypto side only; the
rippled-side recommendations live in that repo.

## Problem

Several `mpt_*` utility wrappers and ElGamal primitives write to
caller-provided output buffers before a downstream check or
operation can still fail. On the failing path, the caller sees a
half-populated output buffer and a failure return code — easy to
misuse, and harder to audit than a strict "compute everything in
temporaries, then commit" pattern.

## Specific sites (line numbers from TOB May 2026; verify by
function name during sweep)

**`mpt-crypto` utility wrappers (`src/utility/mpt_utility.cpp`):**

- `mpt_get_confidential_send_proof` (~L554): calls
  `secp256k1_compact_standard_prove` (which can `return -1`) before
  the `amount > balance_params->amount` check, and before the
  bulletproof generation at `mpt_get_bulletproof_agg` which can
  fail and `return rc`. A failing path can leave the compact-proof
  prefix written to `out_proof`.
- `mpt_get_convert_back_proof` (~L689): writes the compact
  ConvertBack proof to `out_proof` before computing the BP and
  appending it. If BP generation fails, the compact prefix remains
  in the caller buffer.
- `mpt_get_bulletproof_agg` (~L82): may update `*out_len` and write
  to `out_proof` before a later expected-size check fails; on
  failure `*out_len` is left at whatever the prover wrote, not
  restored to the original value.
- `mpt_serialize_ec_pair` (~L255): writes the first point into `out`
  before serializing the second; if the second serialization fails,
  the function returns false with the first 33 bytes already
  written.
- `mpt_generate_keypair` (~L365): calls
  `secp256k1_elgamal_generate_keypair` (which writes
  `out_privkey`) before serializing the public key to `out_pubkey`.
  If pubkey serialization fails, the function returns -1 while
  leaving the private-key output populated.

**ElGamal primitives (`src/elgamal.c`):**

- `secp256k1_elgamal_encrypt` (~L72): writes `c1` before computing
  and writing `c2`. If the shared-secret multiplication returns 0
  or the later point combination returns 0, `c1` is already
  modified.
- `secp256k1_elgamal_add` (~L183): writes `sum_c1` before
[...]

PR #117, opened 2026-06-03
test: add point at infinity regression tests for TOB-RIPCTXR-5

mrtcnk:
## Summary

This PR adds two regression tests to test_elgamal.c that verify the
library's behavior when homomorphic addition produces the point at
infinity — the core cryptographic condition underlying TOB-RIPCTXR-5.

## Tests Added

**test_homomorphic_add_point_at_infinity**
Verifies that secp256k1_elgamal_add correctly returns 0 (failure) when
the resulting C1 component is the point at infinity. This occurs when two
ciphertexts with equal and opposite randomness r and -r are added
homomorphically: C1_result = r*G + (-r)*G = point at infinity.

**test_homomorphic_add_post_mergeinbox_attack**
Simulates a post-MergeInbox inbox locking attack (TOB-RIPCTXR-5 variant).
After MergeInbox, CB_IN resets to canonical zero Enc(0; r0_A) with publicly
known randomness r0_A. A malicious sender can craft a Send with
r_send = -r0_A such that the homomorphic addition of CB_IN and the
malicious Send produces C1 = point at infinity. This test verifies that
secp256k1_elgamal_add returns 0, meaning the malicious Send is rejected
at the library level and CB_IN is never updated with an invalid ciphertext.

## Test Output

Running test: homomorphic add produces point at infinity...
  secp256k1_elgamal_add returned: 0 (expected 0 = failure)
Test passed! Point at infinity correctly rejected.

Running test: post-MergeInbox inbox locking attack (TOB-RIPCTXR-5 variant)...
  Malicious Send ciphertext is valid (C1 != infinity): OK
  Homomorphic add (CB_IN + malicious Send) returned: 0
  Expected: 0 (failure — malicious Send rejected at library level)
Test passed! Malicious Send correctly rejected at library level.
  Our fix combined with TOB-RIPCTXR-14 is sufficient.

## Security Notes

The post-MergeInbox attack is only fully blocked when combined with
correct return value checking of secp256k1_elgamal_add in the rippled
Send state update — as required by TOB-RIPCTXR-14. Both fixes together
are necessary for complete protection.

## Related
- TOB-RIPCTXR-5: Public ciphertext randomness allows attackers to lock inbox funds
- TOB-RIPCTXR-14: Missing return value validation

### DIFF
--- a/tests/test_elgamal.c
+++ b/tests/test_elgamal.c
@@ -14,250 +14,375 @@ static void test_zero_encryption(const secp256k1_context *ctx);
 static void test_canonical_zero(const secp256k1_context *ctx);
 static void test_verify_encryption(const secp256k1_context *ctx);
 static void test_decryption_boundaries(const secp256k1_context *ctx);
+static void test_homomorphic_add_point_at_infinity(const secp256k1_context *ctx);
+static void test_homomorphic_add_post_mergeinbox_attack(const secp256k1_context *ctx);
 
 // Main test runner
 int main(void)
 {
-  secp256k1_context *ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN |
-                                                    SECP256K1_CONTEXT_VERIFY);
-  EXPECT(ctx != NULL);
+    secp256k1_context *ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN |
+                                                      SECP256K1_CONTEXT_VERIFY);
+    EXPECT(ctx != NULL);
 
-  unsigned char seed[32];
-  EXPECT(RAND_bytes(seed, sizeof(seed)) == 1);
-  EXPECT(secp256k1_context_randomize(ctx, seed) == 1);
+    unsigned char seed[32];
+    EXPECT(RAND_bytes(seed, sizeof(seed)) == 1);
+    EXPECT(secp256k1_context_randomize(ctx, seed) == 1);
 
-  test_key_generation(ctx);
-  test_encryption(ctx);
-  test_encryption_decryption_roundtrip(ctx);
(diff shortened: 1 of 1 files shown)


tesseract-ripple (approved):
Rebased onto current `main` and addressed review comments. Pushed directly to the PR branch (maintainer edit).

**Merge conflict resolution:** The conflicts were entirely due to PRs #127/#133 adding the configurable `[range_low, range_high]` API to `secp256k1_elgamal_decrypt` — the PR branch was using the old single-arg signature. Resolved by keeping main's `test_decryption_boundaries` (with the full ranged-API boundary suite) and grafting in the two new test functions.

**Additional fixes applied in the follow-up commit:**

1. **Domain tag coupling** — Extracted `"EncZero"` into a `CANONICAL_ZERO_DOMAIN` macro with a comment pointing to `generate_canonical_encrypted_zero` in `src/elgamal.c`. All offsets in `hash_input` are now derived from `sizeof(CANONICAL_ZERO_DOMAIN) - 1` rather than the magic literal `7`, so a rename in the source will produce a compile-time mismatch rather than a silent wrong-answer test.

2. **Output buffer invariant on failure** — Both new tests now zero-initialize `result_c1`/`result_c2` before the failing `secp256k1_elgamal_add` call and assert the buffers are unmodified after the failure return. This is a regression guard for issue #115 (partial state changes before failure).

3. **Misleading printf removed** — The unconditional `"If 1: attack succeeds — CB_IN would store point at infinity"` line printed even when the test passed. Moved the explanation into the block comment above the call; the test now only prints on success.

Also added `<openssl/evp.h>` explicitly (required for `EVP_Digest` in the attack test).

PR #118, opened 2026-06-08
fix: CI error treating warning as errors

PeterChen13579:


mathbunnyru (approved):
The fix for conanfile.py should also be appled to https://github.com/XRPLF/conan-center-index/blob/master/recipes/mpt-crypto/all/conanfile.py on next release

PR #125, opened 2026-06-15
fix variable collision with rippled and goto statements

PeterChen13579:


yinyiqian1:
The cleanseGuard refactor looks great!
Can we also apply to the  ContextHolder's seed?

tesseract-ripple:
this looks great - we had a number of audit report issues related to cleansing secret variables, so having a more systematic approach will be very helpful. approved pending resolving CI issues.

tesseract-ripple (approved):

PR #126, opened 2026-06-15
fix: reduce MPT_ELGAMAL_DECRYPT_MAX_RANGE to 2000 for practical performance

mrtcnk:
Reduce MPT_ELGAMAL_DECRYPT_MAX_RANGE to 2000 for practical perfrmance

The brute-force DLP solver was iterating up to 1,000,000 steps per decrypt call, causing unacceptable latency in rippled tests. Reduce to 2,000 to unblock testing of TOB-RIPCTXR-5 rerandomization fix.

Off-chain callers requiring larger ranges should use secp256k1_elgamal_decrypt_bsgs() (feature/bsgs-dlp-integration).

Fixes: rippled PR #7535 test blocking issue.


tesseract-ripple (approved):
This constraint should be fine for testing etc before the amendment proposal.

PR #127, opened 2026-06-18
feat(elgamal): add configurable [range_low, range_high] to secp256k1_elgamal_decrypt

mrtcnk:
Summary
Replaces the hardcoded MPT_ELGAMAL_DECRYPT_MAX_RANGE constant with a configurable [range_low, range_high] parameter pair in secp256k1_elgamal_decrypt and mpt_decrypt_amount. This allows callers to specify the decryption range at call time rather than compile time.

Changes

- Removed hardcoded #define MPT_ELGAMAL_DECRYPT_MAX_RANGE
- Updated secp256k1_solve_dlp_small_range_fixed to accept range_low and range_high — computes the starting point range_low * G via scalar multiplication when range_low > 1
- Updated secp256k1_elgamal_decrypt to accept and validate range_low and range_high, suppresses zero match if range_low > 0
- Updated mpt_decrypt_amount to accept and pass through range_low and range_high

tests/test_elgamal.c

- Updated all decrypt calls to pass explicit range parameters
- Expanded test_decryption_boundaries to cover default range, custom range [500, 1500], invalid range, and zero exclusion when range_low > 0

tests/test_mpt_utility.cpp

- Updated mpt_decrypt_amount call site to pass [0, 2000]
- Updated TODO comments to reflect parametrized approach

Notes

- The constant-time fixed-iteration property is preserved within the specified [range_low, range_high] interval
- Zero (m=0) is handled separately and is only accepted if range_low == 0
- BSGS integration will follow in a subsequent PR after internal and external review

### DIFF
--- a/include/utility/mpt_utility.h
+++ b/include/utility/mpt_utility.h
@@ -174,7 +174,11 @@ mpt_encrypt_amount(
  * @param ciphertext [in]  A 66-byte buffer containing the two points (C1, C2).
  * @param privkey    [in]  The 32-byte private key.
  * @param out_amount [out] Pointer to store the decrypted uint64_t amount.
- * @return 0 on success, -1 on failure.
+ * @param range_low  [in]  Lower bound of the search range (inclusive).
+ * @param range_high [in]  Upper bound of the search range (inclusive).
+ * @return 0 on success, -1 on failure, -2 if range_low > range_high.
+ * @note Performance scales linearly with (range_high - range_low). A range of [0, 1,000,000] takes
+ * approximately 3 seconds on Apple Silicon. Do not pass arbitrarily large ranges.
  */
 int
 mpt_decrypt_amount(
--- a/include/utility/mpt_utility.h
+++ b/include/utility/mpt_utility.h
@@ -180,7 +180,9 @@ int
 mpt_decrypt_amount(
     uint8_t const ciphertext[kMPT_ELGAMAL_TOTAL_SIZE],
     uint8_t const privkey[kMPT_PRIVKEY_SIZE],
-    uint64_t* out_amount);
+    uint64_t* out_amount,
+    uint64_t range_low,
+    uint64_t range_high);
 
 /* ============================================================================
  * ZKProof Generation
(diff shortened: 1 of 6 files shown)


include/utility/mpt_utility.h
```diff
@@ -180,7 +180,9 @@ int
 mpt_decrypt_amount(
```
yinyiqian1, line 184: Maybe we want to add in this API's comment, if the range is too large, it will have performance issue. Otherwise users might assume it is functionally not working

src/utility/mpt_utility.cpp
```diff
@@ -498,7 +498,9 @@ int
 mpt_decrypt_amount(
     uint8_t const in_ciphertext[kMPT_ELGAMAL_TOTAL_SIZE],
     uint8_t const privkey[kMPT_PRIVKEY_SIZE],
-    uint64_t* out_amount)
+    uint64_t* out_amount,
+    uint64_t range_low,
+    uint64_t range_high)
 {
     if (!in_ciphertext || !privkey || !out_amount)
         return -1;
```
aanchal4, line 506: `range_low > range_high` is not validated here - `secp256k1_elgamal_decrypt`
will return 0 and this function will return -1, indistinguishable from any
other failure. Suggest adding an explicit guard with a distinct return code:

    if (range_low > range_high)
        return -2;

And documenting -1 vs -2 in the header.

include/utility/mpt_utility.h
mrtcnk: Good suggestion. Added a note to mpt_decrypt_amount in mpt_utility.h.

src/utility/mpt_utility.cpp
mrtcnk: Thanks! Added an explicit guard in mpt_utility.cpp returning -2 when range_low > range_high, and documented the return codes (0 success, -1 failure, -2 invalid range) in the mpt_utility.h doxygen.

yinyiqian1 (approved):

aanchal4 (approved):
LGTM. now! Thanks.

PR #130, opened 2026-06-26
feat: add Baby-Step Giant-Step DLP solver for large-range ElGamal decryption

mrtcnk:
## Summary

Adds `secp256k1_elgamal_decrypt_bsgs()` — a Baby-Step Giant-Step (BSGS) discrete-logarithm solver that recovers EC-ElGamal plaintexts from a range up to 2^40 (≈ 1 trillion). This complements the existing `secp256k1_elgamal_decrypt()`, which is limited to 1,000,000.

## Changes

- `include/bsgs_dlp.h` — public API: `secp256k1_elgamal_bsgs_ctx`, `secp256k1_elgamal_bsgs_ctx_create()`, `secp256k1_elgamal_bsgs_ctx_destroy()`, `secp256k1_elgamal_decrypt_bsgs()`
- `src/bsgs_dlp.c` — implementation
- `tests/test_bsgs_dlp.c` — unit tests covering the full API
- `conan.lock` — regenerated to pin the updated `secp256k1/0.7.1` recipe (which now exports `field.h`, `group.h`, and related internal headers via `private/`)
- `CMakeLists.txt` — removed temporary `secp256k1_internal/` include path; now uses the `<private/...>` conan convention

## Security notes

- **NOT constant-time.** The giant-step loop exits on match, so decryption time leaks the plaintext magnitude. This solver is intended for off-chain use only (local wallets/tools), not server-side infrastructure.
- The existing `secp256k1_elgamal_decrypt()` is unaffected.
- Both decrypt functions are exposed; callers choose based on expected plaintext range.


tesseract-ripple:
## Correctness bug: plaintexts that are odd multiples of `Mhalf = 2^(l1-1)` are unrecoverable

Reviewing the BSGS index scheme, I believe `secp256k1_elgamal_decrypt_bsgs()` silently fails to recover any plaintext `m` that is an **odd multiple of `Mhalf = 2^(l1-1)`**, returning `0` (not-found) for a perfectly valid in-range value.

### Root cause
The baby table is built for `i ∈ [1, Mhalf)` (`src/bsgs_dlp.c`, both build passes: `for (uint64_t i = 1; i < b->Mhalf; i++)`), but the giant step uses stride `M = 2·Mhalf` together with the negation shortcut, so each giant step `j` covers the interval `[j·M − i, j·M + i]` for baby value `i`. With `i` capped at `Mhalf − 1`, every window falls one short of its own midpoint `j·M + Mhalf`. Consecutive windows therefore meet like this:

```
window j    : ... , j·M + (Mhalf−1)]
window j+1  : [(j+1)·M − (Mhalf−1), ...   =  [j·M + Mhalf + 1, ...
                                   ^^^^^^^^^^^^
                       j·M + Mhalf is in neither window
```

`j·M + Mhalf = (2j+1)·Mhalf`, i.e. the gap is exactly the **odd multiples of `Mhalf`**. The `j=0` baby range `[1, Mhalf)` likewise omits `Mhalf` itself.

### Impact
A silent, data-dependent decryption failure (returns `0`, `*amount` untouched) — not a crash. At the default `l1 = 22`, `Mhalf = 2^21`, so the unrecoverable values are the odd multiples of 2,097,152 (`2097152, 6291456, 10485760, …`). A wallet holding exactly one of these balances could not decrypt it.

### Reproduced at runtime
With the test parameters (`l1 = 15`, `Mhalf = 16384`), encrypting `m ∈ {16384, 49152, 81920, 114688, 147456, 180224}` (odd multiples of `Mhalf`) all return `ok = 0`; even-multiple controls all succeed. The existing test suite passes only because its hand-picked value lists happen to contain no odd multiple of `Mhalf`.

### Fix
Make the baby walk inclusive of `Mhalf` — `for (i = 1; i <= b->Mhalf; i++)` in both passes (and `n = Mhalf`, `i + 1 <= b->Mhalf` on the advance guards). Then each giant window becomes `[j·M − Mhalf, j·M + Mhalf]` and adjacent windows meet exactly at the midpoints. This adds a single baby entry; the resulting 1-wide `±` overlap is harmless since `verify_candidate()` disambiguates. Verified: an exhaustive sweep over `[0, 2^12)` (small `l1=6` context) fails at `m=32` before the fix and passes after, with the rest of the suite still green.

I've added a commit to this PR branch with this fix plus an exhaustive-coverage-sweep regression test (the sweep is the important part — a hand-picked list will keep missing this class of off-by-one).

tesseract-ripple (commented):
Follow-up on the minor items I'd drafted here: all four are now addressed in `810be48b` (pushed to this branch). Recording them with their resolutions rather than requesting changes.

- **Dead include** — `#include "bsgs_dlp.h"` removed from `src/elgamal.c` (nothing referenced it). ✅
- **Stale comment reference** — the `bsgs_dlp_benchmark_cached.c` reference in `src/bsgs_dlp.c` replaced with an accurate description of the helpers. ✅
- **`l1` upper bound** — `secp256k1_elgamal_bsgs_ctx_create` now rejects `l1 > 32`. Baby values `i ∈ [1, 2^(l1-1)]` are stored in a `uint32_t`; `l1 ≥ 34` would overflow that field (and `l1 = 33` sits right at the boundary), so the bound is explicit rather than relying on the multi-GB allocation failing. ✅
- **Direct-hit range guard** — the `i=0` direct-hit path (`m = j·M`) is now guarded by the same `max_m` check the negation candidates already use, so `m = 2^bits_total` at `j = J` is rejected rather than reported as found. In the windowed loop the direct hit resolves and returns immediately (its `Qj` is the point at infinity and must not reach batch inversion). Regression assertion added. ✅

Separately, `77c9d6bb` strengthens the exhaustive sweep to run across windows `{128, 8, 1}` so the windowed `fe_batch_invert_tree` path (previously only the tail loop was exercised) and the single-element inversion tree are both covered.

No further changes requested on these items.

tesseract-ripple (approved):
Approving. The correctness bug (plaintexts that are odd multiples of `Mhalf` were silently unrecoverable) and all of Samuel's audit notes are resolved, and the suite is green under ASAN:

- **Mhalf coverage fix** + an exhaustive `[0, 2^12)` sweep across windows `{128, 8, 1}` (exercises the tail, windowed TreeMon, and single-element-tree paths).
- **`window` validation** and the cache **`stash_count` bounds check** — the latter confirmed to fix a heap-buffer-overflow in `map_get_all` (verified via ASAN: trips without the guard, clean with it).
- Dead include removed, stale comment fixed, `l1 > 32` rejected, and the `i=0` direct-hit path now range-guarded.
- The windowed TreeMon batch inversion was verified numerically against the field prime.

For transparency: the last four commits here were pushed by me (the reviewer) to close out the audit notes, so a second glance from @mrtcnk on those is welcome even though I'm approving.

Non-blocking future hardening (not audit findings): the baby-table cache is host-endian and struct-layout-dependent — fine for the intended local single-machine use, and a mismatched cache fails closed via `verify_candidate` — and `baby_load` could additionally validate `section_size` for consistency with `l1`.

src/bsgs_dlp.c
```diff
@@ -0,0 +1,972 @@
...
+ * Jacobian / affine helpers
+ *
+ * Thin wrappers over libsecp256k1's internal field/group primitives. They
+ * are static — not exported — so there is no symbol conflict with any other
+ * translation unit.
+ * ========================================================================= */
+
+/* Load a secp256k1_pubkey into an affine secp256k1_ge.
+ * Mirrors libsecp256k1's internal pubkey_load(). */
+static void pubkey_to_ge(const secp256k1_pubkey *pk, secp256k1_ge *ge)
```
tesseract-ripple, line 65: Minor / non-blocking: `pubkey_to_ge` reads `pk->data` directly and depends on libsecp256k1's internal `secp256k1_ge_storage` layout (the `sizeof(...) == 64` branch). Correct for the pinned 0.7.1 recipe, but coupled to the vendored internal representation — a `static_assert` on the storage size (or a comment pinning the version assumption) would keep a future secp bump from silently breaking it.

src/bsgs_dlp.c
```diff
@@ -0,0 +1,972 @@
...
+    fclose(f);
+    return 0;
+  }
+
+  /* Reject a header whose stash_count is out of range: it is read from an
+   * untrusted file and later used as the bound of a loop over the fixed-size
+   * stash_xb[CUCKOO_STASH_SZ] / stash_val[CUCKOO_STASH_SZ] arrays, so a value
+   * above CUCKOO_STASH_SZ (or negative) would cause an out-of-bounds read. */
+  if (hdr.magic != BABY_MAGIC || hdr.version != BABY_VERSION ||
+      (int)hdr.l1 != expected_l1 || hdr.section_size == 0 ||
```
tesseract-ripple, line 417: Minor / non-blocking: the header check now bounds `stash_count`, but `section_size` is only rejected when `== 0`. A corrupt-but-allocatable `section_size` inconsistent with `l1` would load a bogus table and cause silent decrypt misses until the cache is rebuilt (never a wrong answer — `verify_candidate` fails closed). Recomputing the expected `section_size` from `l1` and rejecting on mismatch would make the cache fully self-consistent.

src/bsgs_dlp.c
```diff
@@ -0,0 +1,972 @@
...
+  hdr.version = BABY_VERSION;
+  hdr.l1 = (uint32_t)l1;
+  hdr.section_size = (uint64_t)baby->section_size;
+  hdr.used_count = (uint64_t)baby->size;
+  hdr.stash_count = (int32_t)baby->stash_count;
+
+  int ok = (fwrite(&hdr, sizeof(hdr), 1, f) == 1);
+  ok &= (fwrite(baby->stash_xb, sizeof(baby->stash_xb), 1, f) == 1);
+  ok &= (fwrite(baby->stash_val, sizeof(baby->stash_val), 1, f) == 1);
+  ok &= (fwrite(baby->tab, sizeof(entry_packed), baby->total_bins, f) ==
```
tesseract-ripple, line 392: Minor / non-blocking: `key`/`val` are written in host byte order and the format depends on struct layout, so a cache file isn't portable across architectures. Fine for the intended local single-machine use (and a mismatched cache fails closed), but worth a one-line note in the header doc — or a byte-order-normalized format — if cache files might ever be shared between machines.

tesseract-ripple (approved):
Approving (re-affirming the earlier approval). No blockers — all Trail of Bits notes are resolved and the suite is green under ASAN. A few non-blocking suggestions inline for a future hardening pass; none should gate this merge or the follow-up release.

src/bsgs_dlp.c
```diff
@@ -0,0 +1,972 @@
...
+ * Recomputes m*G from scratch and compares to the serialized target.
+ * Called at most once per solve on a true match (or a false positive).
+ * ========================================================================= */
+
+static int verify_candidate(const secp256k1_context *ctx, uint64_t m,
+                            const unsigned char target33[33])
+{
+  if (m == 0)
+    return 0;
+
```
aanchal4, line 491: Nit / non-blocking: this guard is unreachable: every call site excludes m == 0 before calling verify_candidate (the m1/m2 checks in bsgs_solve, and the m == 0 path is handled separately before the solver runs at all). Not a correctness issue, just dead code - worth a comment noting it's defensive/intentionally redundant, or removing it.

tests/test_bsgs_dlp.c
```diff
@@ -0,0 +1,511 @@
...
+    }
+  }
+
+  secp256k1_elgamal_bsgs_ctx_destroy(b);
+  printf("Test passed!\n");
+}
+
+static void test_window_validation(const secp256k1_context *ctx,
+                                   secp256k1_elgamal_bsgs_ctx *bsgs)
+{
```
aanchal4, line 413: Nit: All tested window values are either powers of 2 (1, MPT_BSGS_DEFAULT_WINDOW), 0, or out-of-range. No test exercises a non-power-of-2 in-range value (e.g. 100) to confirm the internal round-up-to-power-of-2 in bsgs_solve (while (w < W) w <<= 1;) behaves correctly. Suggest adding one case to close this gap.

aanchal4 (approved):

src/bsgs_dlp.c
mrtcnk: Added a _Static_assert pinning the secp256k1_pubkey <-> ge_storage layout so a future secp bump fails at compile time.

src/bsgs_dlp.c
mrtcnk: baby_load now recomputes the expected section_size from l1 (same ceil(1.3n/3)+2 formula as the build) and rejects any header that disagrees, so an inconsistent-but-allocatable cache fails to load instead of causing silent misses.

src/bsgs_dlp.c
mrtcnk: Documented that the cache is host-endian / struct-layout dependent and not portable across architectures, in both the public header (near cache_path) and the cache-I/O section.

src/bsgs_dlp.c
mrtcnk: Kept the guard and commented it as intentionally redundant/defensive (every caller already excludes m == 0, and secp256k1_ec_pubkey_create would fail on it regardless).

tests/test_bsgs_dlp.c
mrtcnk: Added a window=100 round-trip case to test_window_validation exercising the internal round-up-to-power-of-2 path (the existing cases only used 1 or powers of 2).

PR #131, opened 2026-06-26
fix(scalar): change secp256k1_mpt_scalar_inverse to return int and reject zero input

mrtcnk:
Summary
Addresses the remaining items from TOB-RIPCTXR-9 that were left partially resolved in PR #70.

Changes

- Changed secp256k1_mpt_scalar_inverse signature from void to int
- Returns 0 when the input scalar is zero (inverse is undefined)
- Returns 1 on success

include/secp256k1_mpt.h

- Updated declaration from void to int
- Added documentation comment explaining return values and zero input behaviour

src/bulletproof_aggregated.c

- Updated all 4 call sites to check the return value and goto cleanup or return 0 on failure

### DIFF
--- a/src/bulletproof_aggregated.c
+++ b/src/bulletproof_aggregated.c
@@ -931,7 +931,8 @@ int secp256k1_bulletproof_run_ipa_prover(
       goto cleanup;
 
     /* 4) u_r^{-1} */
-    secp256k1_mpt_scalar_inverse(u_inv, u_scalar);
+    if (!secp256k1_mpt_scalar_inverse(u_inv, u_scalar))
+      goto cleanup;
     if (!secp256k1_ec_seckey_verify(ctx, u_inv))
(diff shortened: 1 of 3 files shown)


src/bulletproof_aggregated.c
tesseract-ripple, line 2470: perhaps we should have a `goto cleanup` here as well? (the PR body does state that all 4 callsites in this file are modified). It's not semantically necessary, since if `y` is zero then line 2457 should catch it.

tesseract-ripple (approved):
this PR fills in the gaps in addressing TOB-RIPCTXR-11, and everything builds and tests correctly. we might want to add one technically-redundant guard for consistency (see the inline comment). also, we might want to add a test for the correct zero-rejection behavior for additional confidence. neither of these are necessary for correctness, so we can go ahead and merge.

PR #132, opened 2026-06-26
fix(crypto): replace variable-time tweak_mul with constant-time wraper for secret scalars (TOB-RIPCTXR-11)

mrtcnk:
## Summary

Addresses the remaining partially resolved items from TOB-RIPCTXR-11. PR #86 introduced `mpt_ct_pubkey_tweak_mul` and applied it to most secret scalar locations, but three locations were missed. This PR fixes those remaining cases.

### `src/commitments.c`
- Line 223: `rho` (secret Pedersen blinding factor) — replaced `secp256k1_ec_pubkey_tweak_mul` with `mpt_ct_pubkey_tweak_mul`

### `src/bulletproof_aggregated.c`
- Line 206: `s_tmp` in `secp256k1_bulletproof_ipa_msm` — replaced with `mpt_ct_pubkey_tweak_mul`
- Line 538: `cLux` in `secp256k1_bulletproof_ipa_compute_LR` (prover only) — replaced with `mpt_ct_pubkey_tweak_mul`
- Line 565: `cRux` in `secp256k1_bulletproof_ipa_compute_LR` (prover only) — replaced with `mpt_ct_pubkey_tweak_mul`
- Line 2576: `tau_x` (secret Bulletproof blinding scalar) — replaced with `mpt_ct_pubkey_tweak_mul`

### DIFF
--- a/src/bulletproof_aggregated.c
+++ b/src/bulletproof_aggregated.c
@@ -203,7 +203,7 @@ int secp256k1_bulletproof_ipa_msm(const secp256k1_context *ctx,
     }
 
     secp256k1_pubkey term = points[i];
-    if (!secp256k1_ec_pubkey_tweak_mul(ctx, &term, s_tmp))
+    if (!mpt_ct_pubkey_tweak_mul(ctx, &term, s_tmp))
       return 0;
 
     if (!add_term(ctx, &acc, &initialized, &term))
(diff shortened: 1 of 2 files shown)


src/bulletproof_aggregated.c
```diff
@@ -2570,7 +2570,7 @@ int secp256k1_bulletproof_verify_agg(
     if (memcmp(tau_x, zero32, kMPT_SCALAR_SIZE) != 0)
     {
       tauH = *h_generator;
-      if (!secp256k1_ec_pubkey_tweak_mul(ctx, &tauH, tau_x))
+      if (!mpt_ct_pubkey_tweak_mul(ctx, &tauH, tau_x))
```
tesseract-ripple, line 2573: I think this change is unnecessary, since it's inside of a verifier function. It's harmless though.

tesseract-ripple (approved):
This PR fills in some sites we missed in earlier PRs to use constant-time functions with secret scalars and does so correctly, following the already-established pattern. One of these changes seems unnecessary (see inline comment), but since it's harmless, we can go ahead and merge anyway.

PR #135, opened 2026-07-20
ci: Add workflow to build WASM artifact and attach to releases

kuan121:
### Why
`mpt-crypto` is used by XRPL client SDKs/libraries in two ways: native shared libraries (xrpl4j, xrpl-py) and a WebAssembly module (xrpl.js). PR #56 gave the **native** libraries a full CI story — build per platform, run `ctest`, bundle, and attach to the GitHub release. The **WASM** module had no equivalent: no CI build, no tests, and no released artifact.

This PR brings WASM to parity, so JS/TS consumers can pull a reproducibly-built, tested wasm module from tagged releases — the same guarantee xrpl4j/xrpl-py already get for the native libs.

## What it adds
```
build-wasm.yml  (one ubuntu job, called by build.yml)
   │
   ├─ setup emsdk 6.0.2 + Node 22
   ├─ build-wasm.sh   → compiles secp256k1 + OpenSSL + mpt-crypto, then links the
   │                    glue TWICE → emcc_out/mpt_crypto.{js, mjs, wasm}
   │                      • mpt_crypto.js   MODULARIZE CommonJS glue
   │                      • mpt_crypto.mjs  EXPORT_ES6 ES-module glue
   │                      • mpt_crypto.wasm the single shared module both glues wrap
   ├─ test-wasm.sh    → (1) compiles the crypto test suite to wasm, runs it under
   │                    Node via ctest; (2) smoke-tests BOTH shipped glues     ← gate
   └─ bundle + upload → mpt-crypto-wasm-bundle  (js + mjs + wasm)
                                        │
   build.yml `release` job (on tag) ────┘
        needs [build-and-test, native-binaries, wasm]
        → attaches mpt-crypto-wasm-<tag>.tar.gz to the GitHub release (next to the native bundle)
```

WASM is a single portable target, so there's no OS×arch matrix — one wasm32 module runs in every browser and in Node/Deno on any platform.

## Dual glue: one wasm, CommonJS + ES module
The wasm module is consumed by xrpl.js, which ships as a **dual CJS+ESM package**. To serve both without shipping the wasm twice, `build-wasm.sh` links the glue **twice from the identical object files**, both wrapping the **same** `mpt_crypto.wasm`:

| Output | emcc flags | Consumer |
|---|---|---|
| `mpt_crypto.js` | `MODULARIZE` (CommonJS) | Node / Jest (`require`) — reads the `.wasm` off disk |
| `mpt_crypto.mjs` | `-sEXPORT_ES6=1` (ES module) | bundlers / browsers (`import`) — locates the wasm via `new URL('mpt_crypto.wasm', import.meta.url)`, which webpack/rollup turn into an emitted asset |

The ESM glue is what lets Confidential MPT run in the browser: the CommonJS glue statically `require`s `node:fs`/`node:crypto` and can't be bundled, whereas the `new URL(import.meta.url)` form makes bundlers emit the wasm as a fetchable asset. Both glues are byte-for-byte the same ABI over the same module — shipping the trio together keeps the glue and wasm in lockstep.

## File by file
| File | Role |
|---|---|
| `.github/scripts/build-wasm.sh` | **Builds the module.** `emcc` link that statically bundles secp256k1 + a stripped OpenSSL, then emits **both** glues (`mpt_crypto.js` MODULARIZE + `mpt_crypto.mjs` EXPORT_ES6) over the same `mpt_crypto.wasm`. Reads dep versions from `conan.lock`; verifies dependency integrity (pinned secp256k1 commit + OpenSSL tarball SHA-256); forces static secp256k1 (`-DBUILD_SHARED_LIBS=OFF`); exports only the symbols consumers use. |
| `.github/scripts/test-wasm.sh` | **Validates it.** (1) Compiles the C crypto test suite to wasm32 and runs each test under Node via `ctest`, reusing the deps `build-wasm.sh` just built. (2) Smoke-tests **both shipped glues** the way consumers load them (`require` the `.js`, `import` the `.mjs`): instantiate, init the secp256k1 context, then `malloc` + run one export + read `HEAPU8` — so a bad `-s` flag or ABI break fails here, not downstream. |
| `.github/workflows/build-wasm.yml` | **The CI workflow.** Reusable (`workflow_call`), single ubuntu job, no matrix. Runs the two scripts, bundles the output (js + mjs + wasm), uploads `mpt-crypto-wasm-bundle`. |
| `.github/workflows/build.yml` | **Wires it in.** Adds a `wasm` job beside `native-binaries`, adds it to `release.needs`, and attaches the wasm tarball on tag push. |
| `.gitignore` | Ignores the emscripten build outputs (`emcc_build/`, `emcc_out/`). |

## Dependency integrity
The native build gets integrity from Conan (`conan.lock` revisions + source SHA-256). This WASM path fetches secp256k1 (git tag) and OpenSSL (release tarball) directly, so it verifies them explicitly to match that guarantee:
- **secp256k1** — the checked-out commit is verified against a pinned SHA (a git tag is mutable; the commit is not).
- **OpenSSL** — the tarball's SHA-256 is verified against a pinned hash before extraction.

## How it's tested
`test-wasm.sh` runs two layers, both gating the `release` job:
1. **Crypto** — compiles mpt-crypto's C crypto test suite to wasm32 and runs all 11 tests under Node via `ctest`.
2. **Glue / ABI** — loads each shipped glue the consumer way (`require` the `.js`, `import` the `.mjs`) and exercises the marshalling contract end to end (secp256k1 context init + `malloc` / `mpt_generate_blinding_factor` / `HEAPU8`).

So a wasm build whose crypto misbehaves — or whose glue/ABI a bad emcc flag has broken — can't ship.

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -1,0 +1,65 @@
+# ──────────────────────────────────────────────────────────────────────────────
+# Build WASM Module
+#
+# Builds the mpt-crypto C library as a single portable WebAssembly module
+# (mpt_crypto.js glue + mpt_crypto.wasm) for JS/TS consumers (xrpl.js). Unlike
+# the native shared libraries there is NO OS x arch matrix: one wasm32 artifact
+# runs in every browser and in Node/Deno/Bun on any platform/endianness.
+#
+# The build also validates the artifact: .github/scripts/build-wasm.sh compiles
+# the C crypto test suite to wasm and runs it under Node (ctest), so the module
+# is exercised before it is bundled — the WASM analog of what build-shared-lib.sh
+# does with ctest for the native libraries.
+#
+# Invoked via `workflow_call` from build.yml — on every PR and push to main (so
+# the build+test path is exercised), and on tag push (so build.yml can attach the
+# bundle to the release, which JS consumers download).
+#
+# The `build` job uploads `mpt-crypto-wasm-bundle` (mpt_crypto.js + .wasm), which
+# the caller downloads and attaches to the GitHub release.
+# ──────────────────────────────────────────────────────────────────────────────
+
+name: Build WASM Module
+
+on:
+  workflow_call:
+
+jobs:
+  build:
+    name: Build + test WASM module
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set up Emscripten
+        uses: mymindstorm/setup-emsdk@6ab9eb1bda2574c4ddb79809fc9247783eaf9021 # v14
+        with:
+          # Pinned for reproducible wasm output. The emsdk version does not
+          # affect crypto results, only the binary's codegen/size.
+          version: 6.0.2
+
+      - name: Set up Node
+        uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6
+        with:
+          node-version: "22"
+
+      - name: Build WASM module
+        run: ./.github/scripts/build-wasm.sh
+
+      - name: Test WASM module
+        run: ./.github/scripts/test-wasm.sh
+
+      - name: Stage WASM bundle
+        run: |
+          mkdir -p output
+          cp emcc_out/mpt_crypto.js emcc_out/mpt_crypto.wasm output/
+          tar -czf mpt-crypto-wasm.tar.gz -C output .
+
+      - name: Upload bundle
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: mpt-crypto-wasm-bundle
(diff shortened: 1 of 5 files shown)


.github/workflows/build-wasm.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Build WASM module
+        run: ./.github/scripts/build-wasm.sh
+
+      - name: Test WASM module
+        run: ./.github/scripts/test-wasm.sh
+
+      - name: Stage WASM bundle
+        run: |
+          mkdir -p output
+          cp emcc_out/mpt_crypto.js emcc_out/mpt_crypto.wasm output/
```
pdp2121: mpt_crypto.mjs seems to be omitted from the bundle here

kuan121: Good catch! I addressed that in this commi https://github.com/XRPLF/mpt-crypto/pull/135/changes/1f643632a7c56b59c19a865391a9f821fbb2fe37 and confirm that the newly built bundle contains 3 files now.

pdp2121 (approved):

PR #136, opened 2026-08-11
ci: Add a self-contained static archive for Rust & Python consumers

pdp2121:
### Why

mpt-crypto previously shipped a **shared** library only. This PR does two related things:

1. **Adds a self-contained static archive** that two sister SDKs need:
   - **xrpl-rust** — a Rust binary that dynamically links a standalone `.so`/`.dylib`/`.dll` isn't self-contained: copying the binary to a server/container without the library alongside it leads to a loader error at startup. Static linking is the idiomatic `-sys`-crate approach and keeps the binary self-contained.
   - **xrpl-py** — the CFFI extension relies on a Windows `LoadLibrary` preload step (Windows has no rpath), and `create_bulletproof`/`verify_bulletproof` are unavailable on Windows because the DLL doesn't export `secp256k1_ec_pubkey_parse`. Static-linking the archive into the `.pyd` removes the preload step and makes those functions available on Windows too (static linking binds non-exported symbols directly).

2. **Unifies the native-library build.** A single Conan dependency build + one CMake configure now produces **both** the static archive **and** the shared library, replacing the previous two-script split whose main difference was a second, redundant dependency build (which rebuilt OpenSSL from source). Java (JNA) and other shared-library consumers are unaffected — the shared library is produced by the same CMake mechanism as before.

### What it does

A single `.github/scripts/build-native-libs.sh` runs one `conan install` (static secp256k1 + OpenSSL) + one CMake configure and produces, per platform:

- the **self-contained static archive** (`libmpt-crypto.a` / `mpt-crypto-static.lib`) with secp256k1 + OpenSSL/libcrypto merged in, and
- the **shared library** (`libmpt-crypto.{so,dylib,dll}`), as a sibling CMake target from the same objects + deps.

**Symbol scope of the static archive:**

- **secp256k1 stays global** — the Rust/Python bindings call stock `secp256k1_*` through the archive. Safe because the Rust `secp256k1` crate renames its symbols to `rustsecp256k1_v0_*` and CPython doesn't use libsecp256k1.
- **OpenSSL is hidden on Linux** — both xrpl-rust (via `reqwest`/`native-tls`) and CPython link the *system* libcrypto; hiding our statically-linked OpenSSL keeps it from interposing with the system copy in-process (ELF flat namespace), which could otherwise cause an ABI mismatch. macOS/Windows don't need it (Secure Transport/SChannel; two-level/per-module namespaces).

### Verification (in CI)

- **Self-containment link test** — links a program against *only* the static archive + base system libs, exercises keypair/blinding/encrypt/commitment, forces the bulletproof/proof members to link, and runs it. Fails if any dependency didn't fold in.
- **Symbol allowlist** — asserts the only global symbols in the archive are the kept API (`mpt_*` + `secp256k1_*`); any other global (e.g. a leaked OpenSSL symbol) fails CI.
- **Shared library** — confirmed to build and export the C API.
- Emits `mpt-crypto-static.link-libs.txt` per platform — the C++ runtime + OpenSSL OS deps a consumer links alongside the archive.

### Artifacts

Per platform (darwin arm64/x86-64, linux x86-64/aarch64/s390x, win32 x86-64): the static archive, the shared library (+ Windows import lib), and the link-libs manifest — plus shared headers and the WASM bundle in the combined natives tarball.

### Notes for reviewers

- Additive for shared-library consumers — the `.so`/`.dylib`/`.dll` is produced by the same CMake shared-library path (`WINDOWS_EXPORT_ALL_SYMBOLS` on Windows) as before, just from the single build.
- Windows uses the dynamic CRT (`compiler.runtime=dynamic`) to match the Python `.pyd` / Rust MSVC consumers.
- ⚠️ Do **not** statically co-link this bundle with an unrenamed secp256k1 (e.g. rippled), or on macOS with a different explicitly-linked OpenSSL — use the shared library there (see the note in `cmake/BundleStatic.cmake`).


.github/workflows/build-shared-libs.yml
```diff
@@ -1,31 +1,47 @@
...
-# (xrpl4j via JNA, xrpl-py via ctypes, etc.) can load directly at runtime.
+# Builds the mpt-crypto C library for each supported platform in TWO forms:
 #
-# secp256k1 and OpenSSL are linked statically INTO the shared library, so
-# consumers do not need either installed on the target machine. The shared
-# library itself still has the usual OS-level dynamic dependencies (libSystem
-# on macOS, glibc/libstdc++ on Linux, Universal CRT on Windows) — there is no
-# attempt at full portability (no musl, no MT CRT).
+#   1. A SHARED library (.dylib / .so / .dll) that client libraries load at
+#      runtime (xrpl4j via JNA, xrpl-py via ctypes, etc.).
```
kuan121: > xrpl-py via ctypes, etc.

Will xrpl-py use this form?

.github/scripts/build-native-libs.sh
```diff
@@ -0,0 +1,262 @@
...
+  CMAKE_ARGS+=(
+    -G Ninja
+    -DCMAKE_BUILD_TYPE=Release
+  )
+fi
+cmake "${CMAKE_ARGS[@]}"
+
+cmake --build build --config Release
+
+# Tests link against the (thin) static target, validating that the build is sound.
```
kuan121: Before, the tests ran against the shared lib. Now they only run against the static build — the .so/.dylib/.dll is built but not tested. Since we already produce it in the same configure, can we run the tests against both?

.github/workflows/build-shared-libs.yml
pdp2121: Great catch! Fixed in [5789d5e](https://github.com/XRPLF/mpt-crypto/pull/136/commits/5789d5e96d01407d53f172e40c93a97bb7f03ad7)

.github/scripts/build-native-libs.sh
pdp2121: Great call! Fixed in [5789d5e](https://github.com/XRPLF/mpt-crypto/pull/136/commits/5789d5e96d01407d53f172e40c93a97bb7f03ad7)

.github/scripts/build-native-libs.sh
```diff
@@ -0,0 +1,279 @@
...
+# One `conan install` (static secp256k1 + OpenSSL) + one CMake configure produces
+# BOTH:
+#   • the self-contained STATIC bundle (libmpt-crypto.a / mpt-crypto-static.lib,
+#     secp256k1 + OpenSSL merged in; see cmake/BundleStatic.cmake), and
+#   • the SHARED library (libmpt-crypto.{so,dylib,dll}), built as a sibling target
+#     from the same objects + the same static deps.
+#
+# This replaces the old build-static-lib.sh + build-shared-lib.sh split, whose
+# only real difference was a second, redundant Conan dependency build (which
+# rebuilt OpenSSL from source).
```
kuan121: This comment seems unnecessary, since once this PR is merged, we can treat build-shared-lib.sh as if it never existed.

kuan121 (approved):

.github/scripts/build-native-libs.sh
pdp2121: Fixed in [60a22a6](https://github.com/XRPLF/mpt-crypto/pull/136/commits/60a22a694543515aaf22fa446535f24e8a521a32)

